PR #9, opened 2017-06-21
QCIF resolution correction

pnkjvvdy:
As per include/config.h 
#define	QCIF	0	// 176  x 144 	AR:	1,222222222
#define	CIF	1	// 352  x 288	AR:	1,222222222
#define	VGA	2	// 640  x 480	AR:	1,333333333

but here line no 501 has QCIF width is 174. So I replace 174 with 176.


murillo128:
Thanxs!

PR #25, opened 2018-05-25
Update pipmosaic.cpp to resolve blinking issue and update audiomixer.cpp for spelling correction

pnkjvvdy:
In the slot, no 2 blinking observed as UV data overlapped of slot one. When a number of slots are 2 then mosaicWidth and mosaicHeight variable's value changed but mosaicWidthUV not updated which calculated based on mosaicWidth.
Rest of another variable should also be changed but it does not any wrong effect.


murillo128:
Could you sing again the new contribution agreement please?

https://www.clahub.com/agreements/medooze/media-server

PR #46, opened 2018-12-27
disable static thin 

notedit:
static lib with `ar -T` will have error when used on  another machine.

see  https://groups.google.com/forum/#!topic/v8-users/WH_Az7NRG2k


murillo128:
change added to the optimization branch will be merged soon

PR #54, opened 2019-01-20
add c_str

notedit:
fix macos build error


notedit:
fix  https://github.com/medooze/media-server-node/issues/66

PR #59, opened 2019-03-03
add last update time

notedit:
add last update time to DTLSICETransport,  

When we got  stun message/rtcp/dtls data  we update the last update time.

Expose a GetLastUpdateTime method,  let  javascript to determine the interval to check this time.

### DIFF
--- a/include/DTLSICETransport.h
+++ b/include/DTLSICETransport.h
@@ -84,6 +84,9 @@ public:
 	virtual int onData(const ICERemoteCandidate* candidate,const BYTE* data,DWORD size)  override;
 	
 	DWORD GetRTT() const { return rtt; }
+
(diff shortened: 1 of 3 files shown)


notedit:
I can finish the javascript part  after this merged.

murillo128:
I want to refactor the way that the js retrieves the stats/info from the c++ objects to avoid requiring locking between the js thread and the event loop one.

Ideally, the js layer will queue a task for retrieving the stats on the event loop trhead that will trigger a callback on the js side (which we can later wrap in a promise).

I was waiting to move from 0.x to 1.x to introduce it, as this would be a breaking change on the API.

notedit:
ok   waiting for the 1.0 version.

PR #74, opened 2019-06-22
Patch 1

pnkjvvdy:


murillo128:
hi @pnkjvvdy could you explain what is the issue this PR solves?

pnkjvvdy:
Greetings,

In asymmetric/parted mosaic it returns always even height, width,top and
left. In some mosaic, Due to odd value mosaic slot border blinking.

In pip mosaic
Clean API implemented.
Update API is looking some complex so I make it simple.
I also removed two unnecessary APIs which is already implemented into base
class Mosaic.

murillo128:
Thank you very much and sorry for the delay in merging!

PR #75, opened 2019-06-26
Changed rateEstimator.minRate to also be used in upload

fuub:


src/videostream.cpp
```diff
@@ -113,7 +114,8 @@ int VideoStream::SetVideoCodec(VideoCodec::Type codec,int mode,int fps,int bitra
 int VideoStream::SetTemporalBitrateLimit(int estimation)
 {
 	//Set bitrate limit
-	videoBitrateLimit = estimation/1000;
+	videoBitrateLimit = estimation/1000;	
+	videoBitrateLimit = sendMinRate ? std::max(sendMinRate,estimation)/1000 : estimation/1000;
```
murillo128: this is duplicated

src/videostream.cpp
```diff
@@ -74,7 +74,8 @@ int VideoStream::SetVideoCodec(VideoCodec::Type codec,int mode,int fps,int bitra
 	//Fix: Should not be here
 	if (properties.HasProperty("rateEstimator.minRate"))
 		//Set it
-		rtp.SetTemporalMinLimit(properties.GetProperty("rateEstimator.minRate",0));
+		rtp.SetTemporalMinLimit(properties.GetProperty("rateEstimator.minRate",0));		
+                sendMinRate = properties.GetProperty("rateEstimator.minRate",0);
```
murillo128: Is this intended to be within the `if`, also, do not duplicate `property.getProperty` call

fuub: sendMinRate is initialized at 0 in videostream.h, so only if profile has the rateEstimator.minRate it will change it.

src/videostream.cpp
```diff
@@ -74,8 +74,8 @@ int VideoStream::SetVideoCodec(VideoCodec::Type codec,int mode,int fps,int bitra
 	//Fix: Should not be here
 	if (properties.HasProperty("rateEstimator.minRate"))
 		//Set it
-		rtp.SetTemporalMinLimit(properties.GetProperty("rateEstimator.minRate",0));		
-                sendMinRate = properties.GetProperty("rateEstimator.minRate",0);
+		sendMinRate = properties.GetProperty("rateEstimator.minRate",0);
+		rtp.SetTemporalMinLimit(sendMinRate);
```
murillo128: Missing brakets

PR #104, opened 2021-01-02
fix function signature for SWIG compiler

marcelhencke:
The [Media Server for Golang](https://github.com/notedit/media-server-go) has the issue, that the swig compiler can't compile the media server correctly. As a result, the support for pushing media streams was deactivated, which is a major turning point.

I have explained the problem in more detail here:
https://github.com/notedit/media-server-go-demo/issues/30#issuecomment-748706452

### DIFF
--- a/include/media.h
+++ b/include/media.h
@@ -174,7 +174,11 @@ public:
 
 	DWORD GetLength() const			{ return buffer->GetSize();			}
 	DWORD GetMaxMediaLength() const		{ return buffer->GetCapacity();			}
+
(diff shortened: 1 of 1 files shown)


murillo128:
I don't feel that the fix is appropriate as I use it for other project which uses SWIG. 

I will close the PR, but leave the issue open and see if i can find a solution that works in all the scenarios

murillo128:
continuing in the issue here

https://github.com/medooze/media-server/issues/105

PR #113, opened 2022-02-04
Add crc32c support for Linux aarch64

mjzac:
Currently the build fails in Linux arm64 platforms. This PR adds the missing crc32c config file.


murillo128:
Thank you!

murillo128:
I assume that it will require too a PR in `media-server-node`, right?

mjzac:
Yes, I have raised a corresponding PR.

PR #121, opened 2022-07-13
raw frame sending PoC

mildsunrise:
introduces functionality to send UDP messages as raw Ethernet frames directly through the interface, for a restricted subset of network configurations. for now we use the portable (but still Linux only) `AF_PACKET` interface to send frames.

this is the low-level part. the JS layer takes care of pulling most of the required parameters from the OS.

### Implementation details

- introduces a RawTxHelper object, which handles:
   - creating & setting up the socket
   - formatting the frame header, calculating checksums, etc.
   - sending frames

  it's constructed with a series of parameters and then offers a TrySend() method (analogous to EventLoop::Send) which attempts to send the UDP message as a raw frame. TrySend() can return false to signal that the packet should be sent
through the OS as normal.

- adds a EventLoop::SetRawTx() method to install this object in an endpoint (which causes the loop to attempt sending through it).

- adds ClearRawTx() and SetRawTx() methods to RTPBundleTransport which serve as wrappers to expose into JS


include/EventLoop.h
```diff
@@ -82,6 +84,11 @@ class EventLoop : public TimeService
 	void Send(const uint32_t ipAddr, const uint16_t port, Packet&& packet);
 	void Run(const std::chrono::milliseconds &duration = std::chrono::milliseconds::max());
 	
+	void SetRawTx(std::optional<RawTxHelper>&& rawTx) {
```
murillo128, line 87: Why optional and shared ptr? let's use an uniquee ptr instead

include/RawTxHelper.h
```diff
@@ -0,0 +1,49 @@
...
+#include <random>
+#include "Packet.h"
+
+/**
+ * @brief Simple helper to send UDP messages as raw Ethernet/IP/UDP
+ * frames in 'typical' network configurations.
+ *
+ * Based on AF_PACKET (Linux-only), doesn't support PACKET_MMAP yet.
+ */
+class RawTxHelper
```
murillo128, line 15: I would prefer the Helper to only do "helper" functions, i.e. handling the eth and udp header creation

include/RawTxHelper.h
```diff
@@ -0,0 +1,49 @@
...
+	 * try to send a frame using raw TX. if false is returned,
+	 * the packet has to be sent through normal means
+	 */
+	bool TrySend(uint32_t ip, uint16_t port, Packet&& packet);
+
+protected:
+	// AF_PACKET socket
+	int fd = -1;
+	// used to check for link-local traffic
+	uint32_t self_addr;
```
murillo128, line 41: let's use camelCase consistently for class members

src/EventLoop.cpp
```diff
@@ -286,6 +286,10 @@ void EventLoop::Send(const uint32_t ipAddr, const uint16_t port, Packet&& packet
 {
 	TRACE_EVENT("eventloop", "EventLoop::Send", "packet_size", packet.GetSize());
 
+	//Try to send through raw TX helper, if installed
+	if (rawTx && rawTx->TrySend(ipAddr, port, std::move(packet)))
```
murillo128, line 290: Why send it here and not just enque it and use the main loop for sending instead?

src/RawTxHelper.cpp
```diff
@@ -0,0 +1,212 @@
...
+	frame.udp.dst = htons(port);
+	frame.udp.length = htons(sizeof(frame.udp) + packet.GetSize());
+	frame.ip.total_len = htons((sizeof(frame) - sizeof(frame.eth)) + packet.GetSize());
+	std::uniform_int_distribution<uint16_t> u16_distr;
+	frame.ip.identification = u16_distr(rng);
+
+	uint32_t checksum;
+
+	// calculate IP header checksum
+	frame.ip.hdr_checksum = 0;
```
murillo128, line 165: Let's move the checksum to an static member of IpHeader

src/RawTxHelper.cpp
```diff
@@ -0,0 +1,212 @@
...
+	checksum = 0;
+	for (size_t i = 0; i < (sizeof(frame.ip_hws) / sizeof(*frame.ip_hws)); i++)
+		checksum += frame.ip_hws[i];
+	checksum = (checksum & 0xFFFF) + (checksum >> 16);
+	checksum = (checksum & 0xFFFF) + (checksum >> 16);
+	frame.ip.hdr_checksum = ~checksum;
```
murillo128, line 171: frame.ip.hdr_checksum = IpHeader::CalculcateChecksum(frame.ip);

src/RawTxHelper.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+	// calculate UDP checksum
+	union {
```
murillo128, line 174: same here

src/RawTxHelper.cpp
```diff
@@ -0,0 +1,212 @@
...
+	// send frame
+	struct iovec iov [2] = { { &frame, sizeof(frame) }, { packet.GetData(), packet.GetSize() } };
+	struct msghdr msg = { nullptr, 0, iov, sizeof(iov)/sizeof(*iov), nullptr, 0, 0 };
+
+	TRACE_EVENT("eventloop", "RawTxHelper::TrySend::sendmsg");
+	if (sendmsg(fd, &msg, 0) < 0) {
+		if (errno == EWOULDBLOCK || errno == EAGAIN) {
+			TRACE_EVENT("eventloop", "RawTxHelper::TrySend::Dropped");
+		} else {
+			throw std::system_error(std::error_code(errno, std::system_category()), "failed sending frame");
```
murillo128, line 206: let's not throw but return false in error cases

murillo128 (changes requested):
I would prefer to move the socket handling to the EventLoop.

We would need to modify the fds poll set to from [socket,pipe] to [readsocket,writesocket,pipe].

If no raw tx is used readsocket == write socket, and if raw socket is set, replace it with the raw socket

include/EventLoop.h
mildsunrise: that still gets us the same problem. lambdas needs to be copyable (so we can pass as `std::function`), so that prevents you from capturing unique_ptrs inside them :(

src/RawTxHelper.cpp
mildsunrise: this should ideally never happen. for now since this is a PoC, I'd prefer the server to terminate if there's an error (since that could indicate a bug) than silently discarding the error and sending the packet through the normal socket. for production I'm not sure how to handle this... we could uninstall the helper and/or print a warning, maybe. WDYT?

src/EventLoop.cpp
mildsunrise: there's the two reasons I didn't use `sendmmsg` for this (and instead do one sendmsg per frame)... one is that there's a [supposedly better approach](https://www.kernel.org/doc/html/latest/networking/packet_mmap.html) where you map a kernel ringbuffer into memory. packets are then enqueued (here) by placing them on the ringbuffer, and then (at the end of the loop iteration) you do a syscall to let the kernel know. the NIC takes them directly from there. this is as close as we can get to Netmap-style frameworks while remaining portable.

the other is just simplicity, this is a PoC and we may decide to add more functionality so I preferred to leave the optimization for later

src/RawTxHelper.cpp
mildsunrise: I think this comment is a duplicate of https://github.com/medooze/media-server/pull/121/files#r922025073?

include/EventLoop.h
murillo128: Just use initialized capture:
```   
void SetRawTx(std::unique_tr<RawTxHelper>&& rawTx ) {
	Async([this, ptr = std::move(rawTxHelper)](auto &now) { this->rawTx = std::move(ptr); });
}
```

bw, don't use elipsis `...` for lambdas, I don't recall why, but it caused some ugly nasty issues that took me a while to debug.

src/RawTxHelper.cpp
murillo128: for production I typically log a warning and discard the packet silently

include/EventLoop.h
mildsunrise: regardless of how you initialize it, capturing a unique_ptr inside a lambda makes it noncopyable (because unique_ptrs are noncopyable) and we need the lambda to be copyable in order to pass it to `Async()`. see: https://taylorconor.com/blog/noncopyable-lambdas/

src/EventLoop.cpp
mildsunrise: yeah, the mmap approach looks like it fits nicely for us so it's an option to consider.
however I'm not sure the supposed benefits are really there, as I've read people claiming that packets are still copied from the mmap ringbuffer into an skb, so... how's it different than using sendmmsg? :sweat_smile: dunno.

in any case I'd agree with waiting a bit. then we can decide if we go directly to Netmap, to mmap or to sendmmsg, if any.

src/RawTxHelper.cpp
mildsunrise: okay, we'll do that then

include/EventLoop.h
murillo128: i get your point now, it is not about moving the pointer but moving the lambda. I think that could be solved by adding a `ASync`  method that takes a  `strd::function<void>(...)&&` as argument (or maybe not)

murillo128: forget what I have just said.. I need to take some rest.. :)

mildsunrise: the helper is noncopyable, it contains a file descriptor. not sure we can duplicate AF_PACKET sockets

murillo128: let's close this for now

src/RawTxHelper.cpp
mildsunrise: done

src/RawTxHelper.cpp
mildsunrise: done

include/RawTxHelper.h
mildsunrise: done

src/RawTxHelper.cpp
mildsunrise: done

include/RawTxHelper.h
mildsunrise: also done

murillo128:
lgtm, I would do minor naming changes or refactoring of some structs, but it is purely cosmetic. Let's merge it.

include/PacketHeader.h
```diff
@@ -0,0 +1,88 @@
...
+struct PacketHeader {
+	using MacAddr = std::array<uint8_t, 6>;
+
+	/**
+	 * @brief Parse an "xx:xx:xx:xx:xx:xx" string into a MacAddr
+	 */
+	static MacAddr ParseMac(std::string str);
+
+
+	struct CandidateData {
```
murillo128, line 24: Could we find a better name for this? like `ArpRoutingInfo` or similar? maybe good to move outside of the `PacketHeader` too

include/PacketHeader.h
```diff
@@ -0,0 +1,88 @@
...
+#include <array>
+#include <string>
+#include <random>
+#include "Packet.h"
+
+/**
+ * @brief Provides structs and logic to build Ethernet packet headers
+ */
+struct PacketHeader {
+	using MacAddr = std::array<uint8_t, 6>;
```
murillo128, line 16: Should we define it as a global type instead?

include/PacketHeader.h
```diff
@@ -0,0 +1,88 @@
...
+
+	/**
+	 * @brief Calculate and (re)set the UDP checksum
+	 */
+	static void CalculateUdpChecksum(PacketHeader& header, const Packet& payload);
+
+	/**
+	 * @brief Initialize the frame header
+	 */
+	static void InitHeader(PacketHeader& header, MacAddr selfLladdr, uint16_t port);
```
murillo128, line 80: maybe do?

static PacketHeader Create( MacAddr selfLladdr, uint16_t port) instead, c++ will do copy elision.

src/RTPBundleTransport.cpp
```diff
@@ -29,6 +33,62 @@
...
+#include <sys/socket.h>
+#include <linux/if_packet.h>
+#include <fcntl.h>
+
+void RTPBundleTransport::SetRawTx(int32_t ifindex, unsigned int sndbuf, bool skipQdisc, uint32_t selfAddr, uint32_t prefixlen, const std::string& selfLladdr, uint32_t gwAddr, const std::string& gwLladdr, uint16_t port)
+{
+
+	// prepare frame template
+
+	PacketHeader header;
```
murillo128, line 55: PacketHeader header = PacketHeader::Create( PacketHeader::ParseMac(selfLladdr), port); ?

src/RTPBundleTransport.cpp
```diff
@@ -29,6 +33,62 @@
...
+	PacketHeader::InitHeader(header, PacketHeader::ParseMac(selfLladdr), port);
+
+	PacketHeader::CandidateData fallbackData = { selfAddr, PacketHeader::ParseMac(gwLladdr) };
```
murillo128, line 58: fallbackData -> defaultRoute ?

src/RTPBundleTransport.cpp
```diff
@@ -545,7 +605,7 @@ void RTPBundleTransport::OnRead(const int fd, const uint8_t* data, const size_t
 			buffer.SetSize(len);
 
 			//Send response
-			loop.Send(ip,port,std::move(buffer));
+			Send(candidate, std::move(buffer));
```
murillo128, line 608: Much better!

murillo128 (approved):

include/PacketHeader.h
mildsunrise: I tried putting it in different headers, but it created dependency loops. it seems a bit small-ish to put in its own header, too...

as for the name, `ArpRountingInfo` seems a bit misleading to me, is `FlowRoutingInfo` okay?

include/PacketHeader.h
mildsunrise: I'm usually against headers defining other things in the global namespace (other than the one with the same name), but if you're okay with making an exception, we can do that

include/PacketHeader.h
mildsunrise: good point

include/PacketHeader.h
murillo128: I am ok about adding a new .h for the `MacAddress` (less be verbose on naming.. :)

include/PacketHeader.h
```diff
@@ -0,0 +1,71 @@
...
+		UdpHeader udp;
+		// for checksum calculation
+		uint16_t udp_hws [sizeof(UdpHeader) / 2];
+	};
+
+
+	/**
+	 * @brief Calculate and (re)set the IP header checksum
+	 */
+	static void CalculateIpChecksum(PacketHeader& header);
```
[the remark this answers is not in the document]
mildsunrise: I imagine this would apply to `PrepareHeader` below, as well? what should we do for `PrepareHeader`, do we make it a class method or do we have it take + return a PacketHeader?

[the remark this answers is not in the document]
murillo128: class method better I think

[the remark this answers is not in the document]
mildsunrise: okay! now, this method can't be const since the checksum ifself needs to be zero while it's being calculated, do we make it a class method as well or something else?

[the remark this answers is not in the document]
murillo128: we can make it a non-const method, maybe even extracting the internal checskum calculation to a helper funciton

src/EventLoop.cpp
```diff
@@ -652,25 +652,37 @@ void EventLoop::Run(const std::chrono::milliseconds &duration)
...
+					sockaddr_in& to		= tos[len];
+					to.sin_family		= AF_INET;
+					to.sin_addr.s_addr	= htonl(item.ipAddr);
+					to.sin_port		= htons(item.port);
+
+					message.msg_name	= (sockaddr*) & to;
+					message.msg_namelen	= sizeof (to);
+				} else {
+					//Packet header
+					auto candidateData = item.rawTxData ? *item.rawTxData : this->rawTx->fallbackData;
```
murillo128, line 678: auto&

murillo128 (changes requested):

src/PacketHeader.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+void PacketHeader::PrepareHeader(PacketHeader& header, std::mt19937 rng, uint32_t ip, uint16_t port, CandidateData rawTxData, const Packet& payload)
+{
+	memcpy(header.eth.ether_dhost, rawTxData.dstLladdr.data(), 6);
+	header.ip.src = htonl(rawTxData.selfAddr);
+	header.ip.dst = htonl(ip);
+	header.udp.dst = htons(port);
+	header.udp.length = htons(sizeof(header.udp) + payload.GetSize());
+	header.ip.totalLen = htons((sizeof(header) - sizeof(header.eth)) + payload.GetSize());
+	std::uniform_int_distribution<uint16_t> u16_distr;
```
murillo128, line 94: no need to be random, let's use an atomic short instead

```
std::atomic_ushort identification;


void PacketHeader::PrepareHeader(PacketHeader& header, std::mt19937 rng, uint32_t ip, uint16_t port, CandidateData rawTxData, const Packet& payload)
{
        memcpy(header.eth.ether_dhost, rawTxData.dstLladdr.data(), 6);
        header.ip.src = htonl(rawTxData.selfAddr);
        header.ip.dst = htonl(ip);
        header.udp.dst = htons(port);
        header.udp.length = htons(sizeof(header.udp) + payload.GetSize());
        header.ip.totalLen = htons((sizeof(header) - sizeof(header.eth)) + payload.GetSize());
        header.ip.identification = htons(identification++);
        PacketHeader::CalculateIpChecksum(header);
        PacketHeader::CalculateUdpChecksum(header, payload);
}
```

src/PacketHeader.cpp
```diff
@@ -0,0 +1,98 @@
...
+}
+
+
+// checksum is a one's complement, so it's independent of endianness
+
+static void ChecksumUpdateWords(uint32_t& checksum, void* data, size_t nWords) {
+	// memcpy is the only safe way to do type punning in C++ pre C++20. the compiler
+	// should see the memcpy call to an identically sized type and optimize it away
+	uint16_t words [nWords];
+	memcpy(words, data, sizeof(words));
```
mildsunrise, line 35: should move the type punning to every individual item, since apparently optimization of the memcpy isn't guaranteed for dynamically-sized arrays...

mildsunrise: the current code (with gcc 12.1, `-O3` and default march) vectorizes to a 9 instruction loop that processes 8 words at once:

~~~ asm
LOOP:
        movdqu  xmm0, XMMWORD PTR [CURSOR]
        add     CURSOR, 16
        movdqa  xmm3, xmm0
        punpckhwd       xmm0, xmm2
        punpcklwd       xmm3, xmm2
        paddd   xmm0, xmm3
        paddd   xmm1, xmm0
        cmp     END, CURSOR
        jne     LOOP
~~~

so it has asymptotic performance of 9/8 = 1.125 instructions / word.
if we need a bit more performance, we could unroll further to get (up to) 6/8 = 0.75 instructions / word.

Sergio mentioned [this article from igalia](https://blogs.igalia.com/dpino/2018/06/14/fast-checksum-computation/) which suggests working in bigger units, so I should try that as well :)

mildsunrise: I spent some time trying many things and was unable to get anything better than the current code.
Using bigger units as suggested in the article results in equal or worse code, and unrollings produce marginal (worse than expected) results.
Also, under AVX the above loop uses YMM regs and processes 16 words at a time, and unrolling seems to produce *worse* results.

I'll leave the current code with the following fixes:
- fix the type punning by doing it element by element (so the memcpy can be optimized away)
- add `__restrict__` to the `const void*` parameter to make sure compiler can optimize correctly
- make it into a separate `CheckSum` class

include/EventLoop.h
```diff
@@ -105,15 +126,17 @@ class EventLoop : public TimeService
 		{
 		}
 		
-		SendBuffer(uint32_t ipAddr, uint16_t port, Packet&& packet) :
+		SendBuffer(uint32_t ipAddr, uint16_t port, std::optional<PacketHeader::CandidateData> rawTxData, Packet&& packet) :
```
murillo128, line 129: `const std::optional<PacketHeader::CandidateData>& rawTxData` instead?

include/RTPBundleTransport.h
```diff
@@ -53,10 +54,14 @@ class RTPBundleTransport :
 	
 	int GetLocalPort() const { return port; }
 	int AddRemoteCandidate(const std::string& username,const char* ip, WORD port);
+	void SetCandidateRawTxData(const std::string& ip, uint16_t port, uint32_t selfAddr, const std::string& dstLladdr);
 	virtual int Send(const ICERemoteCandidate* candidate,Packet&& buffer) override;
 	
 	virtual void OnRead(const int fd, const uint8_t* data, const size_t size, const uint32_t ip, const uint16_t port) override;
 	
+	void SetRawTx(int32_t ifindex, unsigned int sndbuf, bool skipQdisc, const std::string& selfLladdr, uint32_t fallbackSelfAddr, const std::string& fallbackDstLladdr, uint16_t port);
```
murillo128, line 62: what does de `Ll` in `selfLlAddr` means?

mildsunrise: 'lladdr' means 'link-level address' AFAIK. its named this way throughout the Linux APIs, so I thought I'd be consistent with them. but it's true that 'mac' would be more intuitive, so as you prefer

include/PacketHeader.h
mildsunrise: renamed and moved

src/RTPBundleTransport.cpp
mildsunrise: renamed

include/PacketHeader.h
mildsunrise: renamed

include/PacketHeader.h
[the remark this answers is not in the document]
mildsunrise: made into a non-const method

include/CheckSum.h
```diff
@@ -0,0 +1,38 @@
...
+
+class CheckSum
+{
+public:
+	/** 
+	 * IMPORTANT: only the final call to Calculate (before Finalize) can have
+	 * an odd size, otherwise the result will be incorrect (the odd byte should
+	 * be buffered to be combined with the next one as a word)
+	 */
+	inline void Calculate(const void* __restrict__ data, size_t size)
```
murillo128, line 15: couldn't we use uint8_t*/char* instead of void*?

include/MacAddress.h
```diff
@@ -0,0 +1,16 @@
...
+#include <cstdint>
+#include <array>
+#include <string>
+
+class MacAddress : public std::array<uint8_t, 6> {
+public:
+	/**
+	 * @brief Parse an "xx:xx:xx:xx:xx:xx" string into a MacAddress
+	 */
+	static MacAddress Parse(std::string str);
```
murillo128, line 13: const std::string& str

PR #123, opened 2022-08-04
Use const ref for CandidateData and optional and use simple incremental identifier

murillo128:


src/RTPBundleTransport.cpp
```diff
@@ -79,7 +79,7 @@ void RTPBundleTransport::SetRawTx(int32_t ifindex, unsigned int sndbuf, bool ski
 	if (skipQdisc && setsockopt(fd, SOL_PACKET, PACKET_QDISC_BYPASS, &skipQdiscInt, sizeof(skipQdiscInt)) < 0)
 		throw std::system_error(std::error_code(errno, std::system_category()), "failed setting QDISC_BYPASS");
 
-	loop.Async([=](std::chrono::milliseconds) {
+	loop.Async([=, fd = std::move(fd)](std::chrono::milliseconds) {
```
mildsunrise, line 82: this is a bit misleading, remember it gets copied a few times anyway :) in any case I'll leave it here, no big deal

include/FileDescriptor.h
```diff
@@ -31,11 +31,21 @@ class FileDescriptor {
...
-	FileDescriptor& operator=(const FileDescriptor& other) = delete;
 
+	FileDescriptor(const FileDescriptor& other)
+	{
+		fd = dup(other.fd);
+	}
+	FileDescriptor& operator=(const FileDescriptor& other)
+	{
+		fd = dup(other.fd);
+		return *this;
```
mildsunrise, line 42: this leaks if the object is assigned to itself.
to prevent errors like these it's common to use the [copy and swap idiom](https://stackoverflow.com/questions/3279543/what-is-the-copy-and-swap-idiom), as done in the move `operator=` above:

``` c++
FileDescriptor& operator=(FileDescriptor other)
{
	swap(*this, other);
	return *this;
}
```

include/EventLoop.h
```diff
@@ -76,6 +70,20 @@ class EventLoop : public TimeService
...
 		std::function<void(std::chrono::milliseconds)> callback;
 	};
+
+	struct RawTx
+	{
+		FileDescriptor fd;
+		PacketHeader header;
+		PacketHeader::CandidateData fallbackData;
+
+		RawTx(const FileDescriptor& fd, const PacketHeader& header, const PacketHeader::CandidateData& fallbackData)	:
```
mildsunrise, line 80: nit: remember taking const& here forces copying the parameters

murillo128: i know, but i was not able to pass it by move from the lambda capture

include/FileDescriptor.h
murillo128: change it please

include/EventLoop.h
murillo128: (the only movable object is the fd, the others will  e copied anyway)

src/RTPBundleTransport.cpp
mildsunrise: ah yes, moving std::function also moves target so after this it'll only be copied once

PR #124, opened 2022-08-16
send DTLS close notify when stopping transport

mildsunrise:
`DTLSConnection::Reset()` would have been okay, but it calls `Async()` so not an option. I moved its contents into a new `DTLSConnection::Stop()` method. we can also add `SSL_shutdown()` to `Terminate`, but I'm not sure it's good to send close notify on any destruction of the connection


murillo128:
lgtm

PR #129, opened 2023-01-10
MPEG-TS PSI parsing, PAT / PMT

mildsunrise:


src/mpegts/psi.h
```diff
@@ -0,0 +1,91 @@
...
+	static std::vector<ProgramAssociation> ParsePayloadUnit(BufferReader& reader);
+};
+
+/** PSI table data for a Program Map Table */
+struct ProgramMap
+{
+	static const uint8_t TABLE_ID = 0x02;
+
+	/** data for an Elementary Stream entry in a PMT */
+	struct ES
```
murillo128, line 70: rename to ElementaryStream for consistency with acronym usage?

src/mpegts/psi.cpp
```diff
@@ -0,0 +1,178 @@
...
+	uint16_t pid = reader.Get2();
+	// upper 3 bits are reserved and expected to be ones
+	uint8_t reserved = pid >> 13;
+	pid = pid & ~((~0) << 13);
+	if (reserved != 0b111)
+		throw std::runtime_error("invalid PID reserved bits");
+	return pid;
+}
+
+SyntaxData SyntaxData::Parse(BufferReader& reader, const uint8_t*& data, size_t& dataSize)
```
murillo128, line 21: why by ref?

src/mpegts/psi.cpp
```diff
@@ -0,0 +1,178 @@
...
+}
+
+Table Table::Parse(BufferReader& reader)
+{
+	// read table header
+	if (reader.GetLeft()<3)
+		throw std::runtime_error("Not enough data to read mpegts psi table header");
+	BitReader bitreader(reader.GetData(3), 3);
+
+	uint8_t tableId = bitreader.Get(8);
```
murillo128, line 50: Can we add a comment about the spec of the Table element? like:

```Table ID extension	16	Informational only identifier. The PAT uses this for the transport stream identifier and the PMT uses this for the Program number.
Reserved bits	2	Set to 0x03 (all bits on)
Version number	5	Syntax version number. Incremented when data is changed and wrapped around on overflow for values greater than 32.
Current/next indicator	1	Indicates if data is current in effect or is for future use. If the bit is flagged on, then the data is to be used at the present moment.
Section number	8	This is an index indicating which table this is in a related sequence of tables. The first table starts from 0.
Last section number	8	This indicates which table is the last table in the sequence of tables.
Table data	N*8	Data as defined by the Table Identifier.
CRC32	32	A checksum of the entire table excluding the pointer field, pointer filler bytes and the trailing CRC32.

PR #133, opened 2023-01-16
Add missing fields for AAC audio frames

harryz2000:
Adding missing fields for AAC audio frames

### DIFF
--- a/src/rtmp/rtmppacketizer.cpp
+++ b/src/rtmp/rtmppacketizer.cpp
@@ -236,6 +236,10 @@ std::unique_ptr<AudioFrame> RTMPAACPacketizer::AddFrame(RTMPAudioFrame* audioFra
 	//IF we have aac config
 	if (gotConfig)
 	{
+		frame->SetClockRate(aacSpecificConfig.GetRate());
+		frame->SetNumChannels(aacSpecificConfig.GetChannels());
+		frame->SetDuration(aacSpecificConfig.GetFrameLength());
+
 		//Allocate data
 		frame->AllocateCodecConfig(aacSpecificConfig.GetSize());
 		//Serialize it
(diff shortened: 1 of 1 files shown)


src/rtmp/rtmppacketizer.cpp
```diff
@@ -236,6 +236,10 @@ std::unique_ptr<AudioFrame> RTMPAACPacketizer::AddFrame(RTMPAudioFrame* audioFra
 	//IF we have aac config
 	if (gotConfig)
 	{
+		frame->SetClockRate(aacSpecificConfig.GetRate());
```
murillo128, line 239: ClockRate is the clockrate of the timesamps, not the sampling frequency. In rtmp is always 1000.

harryz2000: As per another comment, I am not sure if we need to keep the clock rate 1000 or use the actual sample rate. I just pushed a change to change the timestamp's unit. However, I temporarily keep the actual sample rate as the clock rate, which result in the mpd file has a `audioSamplingRate="48000"` rather than `audioSamplingRate="1000"`.

I am happy to change to 1000 if you prefer.

src/rtmp/rtmppacketizer.cpp
```diff
@@ -228,14 +228,13 @@ std::unique_ptr<AudioFrame> RTMPAACPacketizer::AddFrame(RTMPAudioFrame* audioFra
...
 	//Set time
 	frame->SetTime(audioFrame->GetTimestamp());
-	//Set clock rate
-	frame->SetClockRate(1000);
-	//Set timestamp
-	frame->SetTimestamp(audioFrame->GetTimestamp());
-	
+
 	//IF we have aac config
 	if (gotConfig)
```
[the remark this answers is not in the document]
harryz2000: That is a good point. Yes. The config should be always be sent before the audio frames.

PR #138, opened 2023-02-17
Changes for video resuming has bad images occasionally

harryz2000:
Testing indicates using the original first picture ID as initial value fixed the video resuming image issue.


murillo128:
lgtm

PR #139, opened 2023-02-27
Simulcast recording A/V desync and video quality issue

harryz2000:
When simulcast is enabled, the recording appears to select a low quality layer and sometimes the timestamp causing video pausing. There is also A/V sync issue.

### DIFF
--- a/include/SimulcastVideoMultiplexer.h
+++ b/include/SimulcastVideoMultiplexer.h
@@ -27,7 +27,6 @@ public:
 		if (layerDimensions.find(ssrc) == layerDimensions.end())
 		{
 			layerDimensions[ssrc] = frame->GetWidth() * frame->GetHeight();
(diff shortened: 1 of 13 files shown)


include/SimulcastVideoMultiplexer.h
```diff
@@ -0,0 +1,112 @@
...
+		{
+			auto f = std::move(queue.front());
+			queue.pop_front();
+			forwardFunc(*f);
+		}
+	}
+
+private:
+
+	template<typename T>
```
murillo128, line 78: use a lambda instead of a template

harryz2000: Do you mean use a lambda in the Push function? like below? What's the benefit of that if so?
```
auto enqueue = [this, &forwardFunc](std::unique_ptr<VideoFrame> frame) 
{
....
}
```

PR #143, opened 2023-03-10
Precheck corrupted vp8 packets

harryz2000:
Corrupted packets would cause recording playback on web browser pause. These change is to precheck the vp8 frame header to drop the corrupted packet before going to next stage.

### DIFF
--- a/src/vp8/vp8.h
+++ b/src/vp8/vp8.h
@@ -1,4 +1,4 @@
-/* 
+/*
  * File:   vp8.h
  * Author: Sergio
  *
@@ -46,14 +46,12 @@ struct VP8PayloadDescriptor
 		temporalLayerIndex = 0;
 		keyIndex = 0;
 	}
-	
+
 	VP8PayloadDescriptor(bool startOfPartition,BYTE partitionIndex)
 	{
 		//Empty
 		extendedControlBitsPresent = 0;
 		nonReferencePicture = 0;
-		startOfPartition = 0;
-		partitionIndex = 0;
 		pictureIdPresent = 0;
 		pictureIdLength = 1;
 		temporalLevelZeroIndexPresent = 0;
@@ -94,7 +92,7 @@ struct VP8PayloadDescriptor
 			return 0;
 		//Start parsing
 		DWORD len = 1;
-		
+
 		//Read first
 		extendedControlBitsPresent	= data[0] >> 7;
 		nonReferencePicture		= data[0] >> 5 & 0x01;
@@ -108,7 +106,7 @@ struct VP8PayloadDescriptor
 			if (size<len+1)
 				//Invalid
 				return 0;
-			
+
 			//Read second
 			pictureIdPresent		= data[1] >> 7;
 			temporalLevelZeroIndexPresent	= data[1] >> 6 & 0x01;
@@ -229,7 +227,7 @@ struct VP8PayloadDescriptor
 				//Inc lenght
 				len+=2;
 			} else {
-				//Set picture id 
+				//Set picture id
 				data[len] = pictureId & 0x7F;
 				//Inc lenght
 				len++;
@@ -289,14 +287,14 @@ struct VP8PayloadHeader
 	WORD height = 0;
 	BYTE horizontalScale = 0;
 	BYTE verticalScale = 0;
-	
+
 	DWORD Parse(const BYTE* data, DWORD size)
 	{
 		//Check size
 		if (size<3)
 			//Invalid
 			return 0;
-		
+
 		//Read comon 3 bytes
 		//   0 1 2 3 4 5 6 7
                 //  +-+-+-+-+-+-+-+-+
@@ -306,7 +304,7 @@ struct VP8PayloadHeader
                 //  +-+-+-+-+-+-+-+-+
                 //  |     Size2     |
                 //  +-+-+-+-+-+-+-+-+
-		firstPartitionSize	= get3(data, 0) >> 5;
+		firstPartitionSize	= get3Reversed(data, 0) >> 5;
 		showFrame		= data[0] >> 4 & 0x01;
 		version			= data[0] >> 1 & 0x07;
 		isKeyFrame		= (data[0] & 0x01) == 0;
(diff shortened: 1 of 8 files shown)


src/vp8/vp8depacketizer.cpp
```diff
@@ -12,6 +12,39 @@
 #include "rtp.h"
 #include "log.h"
 
+#ifndef AV_RL24
+#define AV_RL24(x) ((((const uint8_t*)(x))[2] << 16) | (((const uint8_t*)(x))[1] <<  8) | ((const uint8_t*)(x))[0])
```
murillo128, line 16: use `get3`

src/vp8/vp8depacketizer.cpp
```diff
@@ -12,6 +12,39 @@
...
+#endif
+
+namespace {
+
+static bool CheckFrameHeader(uint8_t* buf, size_t buf_size)
```
murillo128, line 21: Isn't this the VP8PayloadHeder already in vp8.h?

murillo128 (changes requested):

src/vp8/vp8depacketizer.cpp
harryz2000: Oh yes. Didn't know that. Will use that in case we still want to use this method. i.e if the sequence number could be used.

src/vp8/vp8depacketizer.cpp
harryz2000: Removed the code

src/vp8/vp8depacketizer.cpp
harryz2000: Change to check first packet and sequence number. Removed this code.

src/vp8/vp8.h
```diff
@@ -306,7 +306,7 @@ struct VP8PayloadHeader
                 //  +-+-+-+-+-+-+-+-+
                 //  |     Size2     |
                 //  +-+-+-+-+-+-+-+-+
-		firstPartitionSize	= data[0] >> 5;
+		firstPartitionSize	= get3(data, 0) >> 5;
```
murillo128, line 309: My bad `get3` and `AVLR_24` has reverse byte order, maybe add a get3reversed or similar name in  tools.h

src/vp8/vp8depacketizer.h
```diff
@@ -19,8 +19,17 @@ class VP8Depacketizer : public RTPDepacketizer
 	virtual MediaFrame* AddPayload(const BYTE* payload,DWORD payload_len) override;
 	virtual void ResetFrame() override;
 private:
+	enum class State
+	{
+		NONE,
```
murillo128, line 24: Use snake camel for states; `None,Processing,Error`

src/vp8/vp8depacketizer.cpp
```diff
@@ -79,8 +98,31 @@ MediaFrame* VP8Depacketizer::AddPacket(const RTPPacket::shared& packet)
...
+	if (packet->GetMark())
+	{
+		switch (state)
+		{
+		case State::PROCESSING:
+			// No error, return the frame
+			validFrame = &frame;
+			break;
+		case State::NONE:
+			Error("-VP8Depacketizer::AddPacket() | Dropped invalid frame as start packet missed. timestamp: %lld\n", packet->GetTimestamp());
```
murillo128, line 114: Use Warning instead of Error, Error is logged into the console logs by default, so this could be quite spammy on production.

src/vp8/vp8depacketizer.cpp
```diff
@@ -103,13 +145,31 @@ MediaFrame* VP8Depacketizer::AddPayload(const BYTE* payload, DWORD len)
 		Error("-VP8Depacketizer::AddPayload() | Error decoding VP8 payload header\n");
 		return NULL;
 	}
-	
+
+	switch (state)
+	{
+	case State::NONE:
```
murillo128, line 151: add 1 more ident

src/vp8/vp8depacketizer.cpp
```diff
@@ -103,13 +145,31 @@ MediaFrame* VP8Depacketizer::AddPayload(const BYTE* payload, DWORD len)
...
+		if (desc.partitionIndex == 0 && desc.startOfPartition)
+		{
+			state = State::PROCESSING;
+		}
+		break;
+	case State::PROCESSING:
+		if (desc.partitionIndex == 0 && desc.startOfPartition)
```
murillo128, line 158: VP8 support multiple partitions per frame, this should not be an error

src/vp8/vp8depacketizer.cpp
```diff
@@ -64,10 +64,29 @@ MediaFrame* VP8Depacketizer::AddPacket(const RTPPacket::shared& packet)
 	}
 	//Set SSRC
 	frame.SetSSRC(packet->GetSSRC());
-	//Add payload
-	AddPayload(packet->GetMediaData(),packet->GetMediaLength());
+
+	if (currentTimestamp != frame.GetTimestamp())
```
murillo128, line 68: Can we merge this timestamp check with the timestamp check in line 50?

murillo128 (changes requested):

murillo128:
Also, can we add unit tests for this?

src/vp8/vp8.h
harryz2000: Didn't notice either :( . Will fix.

src/vp8/vp8depacketizer.h
harryz2000: OK.

src/vp8/vp8depacketizer.cpp
harryz2000: OK

src/vp8/vp8depacketizer.cpp
harryz2000: OK

src/vp8/vp8depacketizer.cpp
harryz2000: It looks this is too strict. The suggestion from the RFC is the partition index SHOULD (not must) increase for each partition. Will remove.

src/vp8/vp8depacketizer.cpp
harryz2000: Yes!

harryz2000:
> 

Yes. We can add unit tests. Will do.

murillo128:
lgtm

PR #144, opened 2023-03-19
Add a function for saving frame config for external use

harryz2000:
This is to expose the code to save frame config so the H264Depacketizer can be used without calling AddPacket() function.

### DIFF
--- a/src/h264/h264depacketizer.cpp
+++ b/src/h264/h264depacketizer.cpp
@@ -67,7 +67,7 @@ MediaFrame* H264Depacketizer::AddPacket(const RTPPacket::shared& packet)
 	if (!packet->GetMark())
 		return NULL;
 
-	SaveFrameConfig();
+	FinalizeFrame();
 
 	//Return frame
 	return &frame;
@@ -370,7 +370,7 @@ MediaFrame* H264Depacketizer::AddPayload(const BYTE* payload, DWORD payloadLen)
 	return &frame;
 }
 
-void H264Depacketizer::SaveFrameConfig()
+void H264Depacketizer::FinalizeFrame()
 {
 	//Set config size
 	frame.AllocateCodecConfig(config.GetSize());
--- a/src/h264/h264depacketizer.cpp
+++ b/src/h264/h264depacketizer.cpp
@@ -29,7 +29,15 @@ H264Depacketizer::~H264Depacketizer()
 
 void H264Depacketizer::ResetFrame()
 {
-	ResetFrameImpl(true);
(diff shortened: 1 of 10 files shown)


src/h264/h264depacketizer.cpp
```diff
@@ -66,10 +66,9 @@ MediaFrame* H264Depacketizer::AddPacket(const RTPPacket::shared& packet)
 	//If it is last return frame
 	if (!packet->GetMark())
 		return NULL;
-	//Set config size
-	frame.AllocateCodecConfig(config.GetSize());
-	//Serialize
-	config.Serialize(frame.GetCodecConfigData(),frame.GetCodecConfigSize());
+
+	SaveFrameConfig();
```
murillo128, line 70: I would prefer some more generic name like `FinalizeFrame` so we could add it to the  `RTPDepacketizer` interface

src/h264/h264depacketizer.cpp
```diff
@@ -370,3 +365,28 @@ MediaFrame* H264Depacketizer::AddPayload(const BYTE* payload, DWORD payloadLen)
...
+	frame.AllocateCodecConfig(config.GetSize());
+	//Serialize
+	config.Serialize(frame.GetCodecConfigData(),frame.GetCodecConfigSize());
+}
+
+void H264Depacketizer::ResetFrameImpl(bool resetConfig)
+{
+	//Clear packetization info
+	frame.Reset();
+	if (resetConfig)
```
murillo128, line 383: do we need to `resetConfig` now the we are cleaning the SPS/PPS in config when receiving a new one?

harryz2000: I would need to read more standard about SPS/PPS about this. My impression is they wouldn't be sent for each frame. I think the receiver/decoder wouldn't need it for each frame as well. For Single NAL Unit, we could always assume there is only one parameter set. But not sure about STAP and MTAPS, as the reset function is also used for those cases.

murillo128: Note that adding it to the config doesn't meant that they are sent on each frame. SPS/PPS are required to be included in each intra frame so new viewers can start watching the stream when they are received. 
I have seen some encoders sending two SPS per intra frame though (in different nal units), single nal unit vs STAP or fragmented nals don't make any difference (MTAPS are not used in rtc)

harryz2000: I changed the code to use config from current frame if it is available. Otherwise, use the previously applied config. I think this would apply all the cases mentioned.

PR #147, opened 2023-03-29
Fix vp8 lagging due to incorrect pic id

harryz2000:


src/rtp/RTPStreamTransponder.cpp
```diff
@@ -366,79 +366,106 @@ void RTPStreamTransponder::onRTP(const RTPIncomingMediaStream* stream,const RTPP
...
-		DWORD temporalLevelZeroIndex = 0;
+
 		//TODO: this should go into the layer selector??
-		if (codec==VideoCodec::VP8 && packet->vp8PayloadDescriptor)
+		if (codec==VideoCodec::VP8)
 		{
+			DWORD pictureId = 0;
+			DWORD temporalLevelZeroIndex = 0;
+
+			if (!packet->vp8PayloadDescriptor)
```
murillo128, line 389: I think it is better to leave the `codec==VideoCodec::VP8 && packet->vp8PayloadDescriptor` in the top if and just pass through the rtp packet

include/rtp/RTPStreamTransponder.h
```diff
@@ -4,7 +4,7 @@
  * and open the template in the editor.
  */
 
-/* 
+/*
```
murillo128, line 7: can we try to avoid doing whitespace changes? it will break git blame

src/rtp/RTPStreamTransponder.cpp
```diff
@@ -366,79 +366,106 @@ void RTPStreamTransponder::onRTP(const RTPIncomingMediaStream* stream,const RTPP
...
+					temporalLevelZeroIndex = *tl0PicIdx;
+
+					//Update ids
+					lastSrcTl0PicIdx = desc.temporalLevelZeroIndex;
 				}
 			}
-		
+
+			// Set if we need rewrite the picture ID or tl0 index
+			packet->rewitePictureIds = packet->vp8PayloadDescriptor->pictureId != pictureId ||
```
murillo128, line 453: I prefer 

```
if (packet->vp8PayloadDescriptor->pictureId != pictureId ||
    packet->vp8PayloadDescriptor->temporalLevelZeroIndex != temporalLevelZeroIndex)
{
    // Set if we need rewrite the picture ID or tl0 index
   packet->rewitePictureIds = true;
    //Rewrite picture id
   packet->vp8PayloadDescriptor->pictureId = pictureId;
    //Rewrite tl0 index
    packet->vp8PayloadDescriptor->temporalLevelZeroIndex = temporalLevelZeroIndex;
}

File: src/rtp/RTPStreamTransponder.cpp [hunk 4]
```diff
@@ -366,79 +366,106 @@ void RTPStreamTransponder::onRTP(const RTPIncomingMediaStream* stream,const RTPP
...
 			{
-				//Update ids
-				lastPicId = desc.pictureId;
-
-				if (!picId)
-				{
-					picId = desc.pictureId;
-				}
-				else
+				if (desc.pictureId != lastSrcPicId || !picId)
```
murillo128, line 401: how is the code handling increasing the pictId different than the previous one? seems the only change is checking the `!picId`

src/rtp/RTPStreamTransponder.cpp
```diff
@@ -366,79 +366,106 @@ void RTPStreamTransponder::onRTP(const RTPIncomingMediaStream* stream,const RTPP
...
-		//Rewrite pict id
-		bool rewitePictureIds = false;
-		DWORD pictureId = 0;
-		DWORD temporalLevelZeroIndex = 0;
+
 		//TODO: this should go into the layer selector??
-		if (codec==VideoCodec::VP8 && packet->vp8PayloadDescriptor)
+		if (codec==VideoCodec::VP8)
 		{
+			DWORD pictureId = 0;
```
murillo128, line 386: I think we can remove this temporal variables and use the ones on the object itself

murillo128 (changes requested):

src/rtp/RTPStreamTransponder.cpp
harryz2000: Sure

src/rtp/RTPStreamTransponder.cpp
harryz2000: as discussed. The previous code doesn't set the rewrite flag to be true for packets other than the first one of a frame.

src/rtp/RTPStreamTransponder.cpp
harryz2000: Yes. We can.

harryz2000:
Will make the change in the other PR: https://github.com/medooze/media-server/pull/146

src/rtp/RTPStreamTransponder.cpp
harryz2000: Sorry. these temporary values are not from the objects. They are from the class members. I can remove this can use the member value directly.

harryz2000: Actually, I found the temporary values makes the code easier to read. Otherwise, I would have to check whether the optional<> has value. Also keep it in consistent format to temporalLevelZeroIndex, which was set different for base layer and other layers. But I moved the variable close to where it is used.

src/rtp/RTPStreamTransponder.cpp
harryz2000: After moving to VP8LayerSelector, there is no need to check the VP8 codec.

PR #151, opened 2023-05-26
Fix RTMP media packet parsing for some corner cases

harryz2000:
Renamed some local variables to avoid duplicating the member variables.
Also added some debugging messages.


src/rtmp/rtmpmessage.cpp
```diff
@@ -1121,50 +1121,51 @@ void RTMPAudioFrame::Dump()
 
 DWORD RTMPAudioFrame::Parse(BYTE *data,DWORD size)
 {
-	BYTE* buffer = data;
-	DWORD bufferLen = size;
+	BYTE* buf = data;
```
murillo128, line 1124: can we find a better name? easy to mistake with the `buffer` member property

PR #152, opened 2023-05-30
Fixed one valgrind uninitialized buffer error

harryz2000:


src/AudioPipe.cpp
```diff
@@ -210,12 +210,18 @@ int AudioPipe::RecBuffer(SWORD* buffer, DWORD size)
...
 	DWORD len = 0;
 
-	//Calculate total audio length
-	DWORD totalSize = size * numChannels;
-
 	//Bloqueamos
 	pthread_mutex_lock(&mutex);
-
+	
+	if (!playing)
```
murillo128, line 216: I think that the correct fix is 

```
while (!playing)
{
		pthread_cond_wait(&cond, &mutex);
               //If we have been canceled
		if (canceled)
		{
			//Remove flag
			canceled = false;
			//Exit
			Log("AudioPipe: RecBuffer cancelled.\n");
			//End
			goto end;
		}
}

harryz2000: I think you want to block until we can get buffer from it. Sure.

PR #153, opened 2023-06-02
Use the transID from RTMP client when sending status response

harryz2000:


src/rtmp/rtmpconnection.cpp
```diff
@@ -1252,11 +1252,10 @@ void RTMPConnection::onCommand(DWORD streamId,const wchar_t *name,AMFData* obj)
 	SendCommand(streamId,name,new AMFNull(),obj->Clone());
 }
 
-void RTMPConnection::onNetStreamStatus(DWORD streamId,const RTMPNetStatusEventInfo &info,const wchar_t *message)
+void RTMPConnection::onNetStreamStatus(DWORD streamId,QWORD transId,const RTMPNetStatusEventInfo &info,const wchar_t *message)
 {
 	RTMPNetStatusEvent event(info.code,info.level,message);
-	//Send command
-	onCommand(streamId,L"onStatus",&event);
+	SendCommandResponse(streamId,L"onStatus",transId,new AMFNull(),event.Clone());
```
murillo128, line 1258: why change onCommand by SendCommandResponse ?

harryz2000: The onCommand is sending command as initiator and not using the transID from the client. SendCommandResponse is to respond a command send from the client, therefore need the transID from the received command.

PR #158, opened 2023-06-26
obsolete: H265 recording for SRT

ZitaLiao:


murillo128:
given that we have committed the HEVCDescriptor in another PR, is this PR still valid, or do we need to create a different one without the changes already commited?

PR #159, opened 2023-06-27
HEVCDescriptor for H265 recording

ZitaLiao:


murillo128:
lgtm, only a few identation with spaces that should be tab. Would it be possible to add some basic unit tests as well?

PR #161, opened 2023-06-28
H.26x packetizer fixes, deduplication

mildsunrise:
this PR extracts the packetization code from Zita's branch in srt-server-node, into an `H265Packetizer` class, and does the same for the H.264 packetizer.

it the factors out the logic common to both (Annex B slicing, fragmentation, insertion of NALUs into the destination frame) into a common base class called `H26xPacketizer`.

~~draft! **please don't merge until we've merged the other H.265 PRs**~~

### DIFF
--- a/src/h264/H264Packetizer.h
+++ b/src/h264/H264Packetizer.h
@@ -1,0 +1,27 @@
+#ifndef H264PACKETIZER_H
+#define H264PACKETIZER_H
+
+#include "video.h"
+#include "avcdescriptor.h"
+
+/**
+ * RTP packetizer for H.264 streams.
+ * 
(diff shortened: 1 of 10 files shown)


src/h264/H264Packetizer.h
```diff
@@ -0,0 +1,39 @@
...
+	AVCDescriptor config;
+
+	std::unique_ptr<VideoFrame> ProcessAU(BufferReader reader) override;
+
+protected:
+	void OnNal(VideoFrame& frame, BufferReader& nal) override;
+
+	void EmitNal(VideoFrame& frame, BufferReader nal)
+	{
+		uint8_t naluHeader 	= nal.Peek1();
```
murillo128, line 28: move to .cpp better

PR #166, opened 2023-07-03
[draft] add reference timestamp

harryz2000:


murillo128:
I preffer the other approach better

PR #167, opened 2023-07-03
Allow adding delay for frame dispatching

harryz2000:


src/MediaFrameListenerBridge.cpp
```diff
@@ -291,13 +291,24 @@ void MediaFrameListenerBridge::onMediaFrame(DWORD ignored, const MediaFrame& fra
 			//UltraDebug("-MediaFrameListenerBridge::onMediaFrame() [this:%p,extSeqNum:%d,pending:%d,duration:%dms,total:%d,total:%dms\n", extSeqNum-1, pendingLength, packetDuration, info[i].GetTotalLength(), packetDuration);
 
 			//Insert it
-			packets.emplace(packet,packetDuration);
+			auto schedued = now + dispatchingDelayMs.load();
+			pendingPackets.emplace(schedued, std::pair<RTPPacket::shared, std::chrono::milliseconds>(packet,packetDuration));
```
murillo128, line 295: I would prefer to enqueue directly to the packets and move the the code checing scheduled vs now to go inside the dispatch timer itself

src/MediaFrameListenerBridge.cpp
```diff
@@ -291,13 +291,24 @@ void MediaFrameListenerBridge::onMediaFrame(DWORD ignored, const MediaFrame& fra
...
 
 			//Recalcualte pending
 			pendingLength -= info[i].GetTotalLength();
 			pendingDuration -= packetDuration;
 
 		}
+		
+		// Check and enque packets that needs to be dispatched.
+		while (!pendingPackets.empty())
```
murillo128, line 304: move this to the dispatch timer

src/MediaFrameListenerBridge.cpp
```diff
@@ -291,13 +291,24 @@ void MediaFrameListenerBridge::onMediaFrame(DWORD ignored, const MediaFrame& fra
 			//UltraDebug("-MediaFrameListenerBridge::onMediaFrame() [this:%p,extSeqNum:%d,pending:%d,duration:%dms,total:%d,total:%dms\n", extSeqNum-1, pendingLength, packetDuration, info[i].GetTotalLength(), packetDuration);
 
 			//Insert it
-			packets.emplace(packet,packetDuration);
+			auto schedued = now + dispatchingDelayMs.load();
```
murillo128, line 294: we need to make sure that the frames are dispatched in order, so in case the dispatchingDelay is lowered we don't have out of order packets.

include/MediaFrameListenerBridge.h
```diff
@@ -31,6 +31,8 @@ class MediaFrameListenerBridge :
 	void Update(QWORD now);
 	void UpdateAsync(std::function<void(std::chrono::milliseconds)> callback);
 	void Stop();
+	
+	void SetDelayMs(std::chrono::milliseconds delayMs) { dispatchingDelayMs = delayMs; };
```
murillo128, line 35: in stead of using atomic what we typically do is call timeService.Async() to set the delay. If the setDelay is happening on the timeService thread, there will cause no overhead and we will skip the atomic load in each media frame

murillo128 (changes requested):

src/MediaFrameListenerBridge.cpp
harryz2000: Yes. Packets are put in the queue. Front packets would be always be dispatched earlier even they have a larger delay.

src/MediaFrameListenerBridge.cpp
harryz2000: Done

src/MediaFrameListenerBridge.cpp
harryz2000: Done.

include/MediaFrameListenerBridge.h
harryz2000: Done.

src/MediaFrameListenerBridge.cpp
```diff
@@ -291,7 +314,8 @@ void MediaFrameListenerBridge::onMediaFrame(DWORD ignored, const MediaFrame& fra
 			//UltraDebug("-MediaFrameListenerBridge::onMediaFrame() [this:%p,extSeqNum:%d,pending:%d,duration:%dms,total:%d,total:%dms\n", extSeqNum-1, pendingLength, packetDuration, info[i].GetTotalLength(), packetDuration);
 
 			//Insert it
-			packets.emplace(packet,packetDuration);
+			auto schedued = now + dispatchingDelayMs;
```
murillo128, line 317: we should ensure that scheduled is increasing always, something like:

```
if (!packets.empty())
    scheduled = std::max(scheduled,packets.front().first+1);
```

murillo128 (changes requested):

src/MediaFrameListenerBridge.cpp
harryz2000: Sure. Will do

harryz2000: Done. Assume you mean the packets.back(). Also move it out of the loop.

include/MediaFrameListenerBridge.h
```diff
@@ -58,7 +60,7 @@ class MediaFrameListenerBridge :
 	TimeService& timeService;
 	Timer::shared dispatchTimer;
 
-	std::queue<std::pair<RTPPacket::shared, std::chrono::milliseconds>> packets;
+	std::queue<std::pair<std::chrono::milliseconds, std::pair<RTPPacket::shared, std::chrono::milliseconds>>> packets;
```
murillo128, line 63: should we use a tuple instead? or maybe create a struct to avoid confusions.

src/MediaFrameListenerBridge.cpp
```diff
@@ -234,6 +257,14 @@ void MediaFrameListenerBridge::onMediaFrame(DWORD ignored, const MediaFrame& fra
 		//Calculate each packet duration
 		uint32_t pendingDuration = frame->GetDuration() * 1000 / rate;
 
+		// Calculate the scheduled time
+		auto schedued = now + dispatchingDelayMs;
```
murillo128, line 261: typo?

src/MediaFrameListenerBridge.cpp
```diff
@@ -106,8 +126,11 @@ void MediaFrameListenerBridge::onMediaFrame(DWORD ignored, const MediaFrame& fra
 		//Remove all packets
 		while (packets.size())
 		{
```
murillo128, line 128: do we really need this loop here now? I was emptying the queue when a new frame was received, but now that we are doing dispatching, we should probably move this to the dispatch method

src/MediaFrameListenerBridge.cpp
```diff
@@ -52,9 +55,19 @@ MediaFrameListenerBridge::MediaFrameListenerBridge(TimeService& timeService,DWOR
...
+		if (packets.size())
+		{
 			//Reschedule
-			dispatchTimer->Again(std::chrono::milliseconds(accumulated - period));
+			auto deferMs = packets.front().first - now;
+			if (deferMs.count() > 0)
+			{
+				dispatchTimer->Again(deferMs);
+			}
+			else if (accumulated > period)
```
murillo128, line 66: is this condition ever meet?

include/MediaFrameListenerBridge.h
harryz2000: Sure. I will create a struct for it.

src/MediaFrameListenerBridge.cpp
harryz2000: Yes. Will fix

murillo128 (approved):

PR #170, opened 2023-07-05
Add FrameDispatchCoordinator class and tests

harryz2000:


src/FrameDispatchCoordinator.cpp
```diff
@@ -0,0 +1,16 @@
...
+					int aUpdateRefsPacketLateThresholdMs, 
+					std::chrono::milliseconds aUpdateRefsStepPacketEarlyMs) :
+	frameDelayCalculator(aUpdateRefsPacketEarlyThresholdMs, aUpdateRefsPacketLateThresholdMs, aUpdateRefsStepPacketEarlyMs)
+{
+}
+
+void FrameDispatchCoordinator::OnFrame(std::chrono::milliseconds now, uint64_t ts, uint64_t clockRate, MediaFrameListenerBridge& listenerBridge)
+{	
+	// Use the address of the MediaFrameListenerBridge as stream identifier
+	auto delayMs = frameDelayCalculator.OnFrame(uint64_t(&listenerBridge), now, ts, clockRate);
```
murillo128, line 13: use `listenerBridge.GetMediaSSRC()` we will need to make sure that the ssrcs are unique when constructing them on the SRT server

test/unit/TestFrameDispatchCoordinator.cpp
```diff
@@ -0,0 +1,2484 @@
+#include <gtest/gtest.h>
+
+#include "FrameDispatchCoordinator.h"
```
murillo128, line 3: can we change the test to be for the FrameDelayCalculator instead?

murillo128 (changes requested):

src/FrameDispatchCoordinator.cpp
harryz2000: Sure.

test/unit/TestFrameDispatchCoordinator.cpp
harryz2000: Yes. The current tests are actually for both FrameDispatchCoordinator (include FrameDelayCalculator)  and MediaFrameListenerBridge dispatching.  I will do a refactor to split some tests for FrameDelayCalculator and keep the tests for MediaFrameListenerBridge in the original test class.

harryz2000: Done.

src/FrameDispatchCoordinator.cpp
harryz2000: Done. The SrtConnection creates the audio and video listener bridge respectively with ssrc 1 and 2.

PR #175, opened 2023-07-17
RTMP V1 implementation

harryz2000:


include/rtmp/rtmpmessage.h
```diff
@@ -71,22 +104,98 @@ class RTMPVideoFrame : public RTMPMediaFrame
...
+		default:
+			return VideoCodec::UNKNOWN;
+		}
+	}
+	
+private:
+	
+	enum class ParsingState
+	{
+		VIDEO_TAG_HEADER,
```
murillo128, line 179: Please use camel case

src/rtmp/rtmppacketizer.cpp
```diff
@@ -1,18 +1,18 @@
 
 #include "rtmp/rtmppacketizer.h"
-#include "h264/h264.h"
 
-std::unique_ptr<VideoFrame> RTMPAVCPacketizer::AddFrame(RTMPVideoFrame* videoFrame)
+template<typename DescClass, typename SPSClass, VideoCodec::Type codec>
```
murillo128, line 4: I don't think this will work when we add support to AV1 as well, maybe it is easier to have an interface for the packetizer and derive each codec paketizer from it instead

include/rtmp/rtmpmessage.h
```diff
@@ -55,12 +74,26 @@ class RTMPMediaFrame
...
 };
 
+
 class RTMPVideoFrame : public RTMPMediaFrame
 {
 public:
-	enum VideoCodec {FLV1=2,SV=3,VP6=4,VP6A=5,SV2=6,AVC=7};
-	enum FrameType  {INTRA=1,INTER=2,DISPOSABLE_INTER=3,GENERATED_KEY_FRAME=4,VIDEO_INFO=5};
-	enum AVCType	{AVCHEADER = 0, AVCNALU = 1, AVCEND = 2 };
+	enum class RtmpVideoCodec {FLV1=2,SV=3,VP6=4,VP6A=5,SV2=6,AVC=7};
```
murillo128, line 81: If it is not strictly necessary, revert the `enum class` to `enum` and the `RtmpVideoCodec` to `VideoCodec`.

While I acknowledge that current behavior is better, we have not changed the Audio classes, and there would be a completely mismatch between audio and video classes.

murillo128 (changes requested):

include/rtmp/rtmpmessage.h
harryz2000: sure

src/rtmp/rtmppacketizer.cpp
harryz2000: No. It wouldn't support AV1 I think. Will have a further look.

include/rtmp/rtmpmessage.h
harryz2000: Just worried about the same value representing different types. But should not be a problem to revert them to enum. Will do.

include/rtmp/rtmpmessage.h
```diff
@@ -71,22 +104,98 @@ class RTMPVideoFrame : public RTMPMediaFrame
...
+		if (!isExtended)
+		{
+			return  GetAVCType() == RTMPVideoFrame::AVCType::AVCNALU;
+		}
+		
+		return GetPacketType() == RTMPVideoFrame::PacketType::CodedFrames ||
+			GetPacketType() == RTMPVideoFrame::PacketType::CodedFramesX;
+	}
+	
+	VideoCodec::Type GetGenericVideoCodec() const
```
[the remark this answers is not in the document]
harryz2000: This is for the packetizer to check the RTMPVideoFrame's codec, which is using VideoCodec enum defined in codecs.h .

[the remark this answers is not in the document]
harryz2000: OK

include/rtmp/objectparser.h
```diff
@@ -0,0 +1,73 @@
...
+	/**
+	 * Parse a chunk of data
+	 * 
+	 * @param data The data to be parsed.
+	 * @param len Length of the data
+	 * 
+	 * @return An pair. The first element is whether the object has been parsed. The second
+	 *         element is the length of data that was used.
+	 */
+	std::pair<bool, size_t> Parse(uint8_t* data, size_t len)
```
murillo128, line 42: This `Parse` method doesn't parse anything, what does it do?

src/rtmp/rtmpmessage.cpp
```diff
@@ -3,6 +3,7 @@
 #include "avcdescriptor.h"
 #include <stdexcept>
 #include <cstdlib>
+#include <cassert>
```
murillo128, line 6: remove include

include/rtmp/objectparser.h
murillo128: Seems that this class is just writting to a buffer, use https://github.com/medooze/libdatachannels/blob/master/src/internal/BufferWritter.h instead

murillo128 (changes requested):

include/rtmp/objectparser.h
harryz2000: This is to make it easy to parse chunked data for a object (currently fixed-size, could be extended to support variable-size object as well). 

You could see it is not a complicated class and it doesn't look to me the BufferWritter has the facility to try to fill as much as possible data, which means we would need to check its capacity and current data size to put in the buffer. But using ObjectParser is make it easier as if data is enough for the object, no data would be put in.

The initial thought was the ObjectParser would be a base class for different parsers, such as SI24Parser, ExtraDataParser, etc. However, I found probably for now it is enough just use the base class (to reduce code). But I kept the name. For example:

```
class SI24Parser : public ObjectParser
{
public:
      SI24Parser() : ObjectParser(3) {}

      uint32_t GetValue() { return (int32_t(get3(buffer.data(), 0)) << 8) >> 8; }
}
```

murillo128 (changes requested):

src/rtmp/rtmpmessage.cpp
```diff
@@ -980,7 +981,9 @@ DWORD RTMPVideoFrame::Parse(BYTE *data,DWORD size)
 			{
 				packetType = PacketType(buffer[0] & 0x0f);
 				
-				objectParser = std::make_unique<ObjectParser>(4);
+				dataBuffer.resize(4);
+				bufferWritter = std::make_unique<BufferWritter>(dataBuffer.data(), dataBuffer.size());
```
murillo128, line 985: can't we use `extraData` directly?

src/rtmp/rtmpmessage.cpp
```diff
@@ -992,35 +995,39 @@ DWORD RTMPVideoFrame::Parse(BYTE *data,DWORD size)
...
-			std::tie(isParsed, usedBytes) = objectParser->Parse(buffer, bufferLen);
-			if (isParsed)
+			usedBytes = std::min(uint32_t(bufferWritter->GetLeft()), bufferLen);
+			(void)bufferWritter->Set({buffer, usedBytes});
+			
+			if (bufferWritter->GetLeft() == 0)
 			{
-				codecEx = VideoCodecEx(FourCcToUint32(objectParser->GetBuffer().begin()));				
-				objectParser.reset();				
+				codecEx = VideoCodecEx(FourCcToUint32(dataBuffer.begin()));
```
murillo128, line 1029: let's add a new `BYTE fourCC[4];` member and use the memory directly

murillo128 (changes requested):

src/rtmp/rtmpmessage.cpp
harryz2000: Done.

include/avcdescriptor.h
```diff
@@ -25,6 +25,11 @@ class AVCDescriptor
 	bool Parse(const BYTE* data,DWORD size);
 	void Dump() const;
 
+	// For compatibility with RTMP packetizer. Not used for AVC
+	BYTE GetNumOfVideoParameterSets()		const { return 0;	}
```
murillo128, line 29: remove (see later comment)

murillo128 (changes requested):

src/rtmp/rtmppacketizer.cpp
```diff
@@ -68,6 +98,22 @@ std::unique_ptr<VideoFrame> RTMPAVCPacketizer::AddFrame(RTMPVideoFrame* videoFra
 	//If is an intra
 	if (videoFrame->GetFrameType()==RTMPVideoFrame::INTRA)
 	{
+		//Add VPS
+		for (int i=0;i<desc.GetNumOfVideoParameterSets();i++)
```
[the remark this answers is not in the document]
harryz2000: How do you think to check the type? Using std::enable_if?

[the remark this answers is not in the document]
murillo128: std::is_same ? https://en.cppreference.com/w/cpp/types/is_same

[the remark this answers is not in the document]
harryz2000: I am not sure std::is_same could work directly for this case, as we have to prevent the code compile when the type is not expected. I think we might have to add override functions such as below. Are you OK with this or do you have other thoughts?

```
void AddVps(DWORD nalUnitLength, BYTE* nalHeader, const AVCDescriptor&desc, VideoFrame* frame)
{
}

void AddVps(DWORD nalUnitLength, BYTE* nalHeader, const HEVCDescriptor&desc, VideoFrame* frame)
{
	for (int i=0;i<desc.GetNumOfVideoParameterSets();i++)
	{
		//Set size
		setN(nalUnitLength, nalHeader, 0, desc.GetVideoParameterSetSize(i));
		
		//Append nal size header
		frame->AppendMedia(nalHeader, nalUnitLength);
		
		//Append nal
		auto ini = frame->AppendMedia(desc.GetVideoParameterSet(i),desc.GetVideoParameterSetSize(i));
		
		//Crete rtp packet
		frame->AddRtpPacket(ini,desc.GetVideoParameterSetSize(i),nullptr,0);
	}	
}
```

murillo128 (approved):

PR #176, opened 2023-07-17
Additional change to make dump and serialisation work

harryz2000:


murillo128:
What is the purpose of this change?

harryz2000:
It is to complete the dump() and serialization() functions for the RTMP v1 message. There is a small refactor as well to unify the new and old RTMP versions with regard to VideoCodec and packet type. Also the compositionTime. Some information are mutually exclusive between two versions.

So I think using variant would make it more organised. Also the extraData is kind of vague, so I changed to get the exact meaning.

murillo128 (approved):
lgtm

PR #179, opened 2023-07-21
[WIP] RTMP + Packetization refactor

harryz2000:


include/H26xNaluPacketizer.h
```diff
@@ -0,0 +1,40 @@
+#ifndef H26XNALUPACKETIZER_H
+#define H26XNALUPACKETIZER_H
```
ZitaLiao, line 2: Currently the H26x files are in src/h264. We may make them together in a single place by the way with this refactor?

src/h264/H264NaluPacketizer.h
```diff
@@ -0,0 +1,42 @@
+#ifndef H264NALUPACKETIZER_H
```
ZitaLiao, line 1: Not sure we need this single function file H26x/H264/H265NaluPacketizer.h? May be we can consider to  put it into h264.cpp and h265.cpp, since Nalu is also part of the codec standard in H264/H265?

src/h265/H265NaluPacketizer.h
```diff
@@ -0,0 +1,40 @@
...
+		auto naluHeader = get2(data, 0);
+		BYTE nalUnitType		= (naluHeader >> 9) & 0b111111;
+		BYTE nuh_layer_id		= (naluHeader >> 3) & 0b111111;
+		BYTE nuh_temporal_id_plus1	= naluHeader & 0b111;
+
+		const uint16_t nalHeaderFU = ((uint16_t)(HEVC_RTP_NALU_Type::UNSPEC49_FU) << 9)
+							| ((uint16_t)(nuh_layer_id) << 3)
+							| ((uint16_t)(nuh_temporal_id_plus1)); 
+		
+		std::vector<uint8_t> fragmentationHeader(3);
```
ZitaLiao, line 31: minor: use HEVCParams::RTP_NAL_HEADER_SIZE and HEVCParams::RTP_FU_HEADER_SIZE instead of hardcode 2/3 to easier understand?

include/H26xFrameAppender.h
```diff
@@ -0,0 +1,53 @@
...
+		frame.AppendMedia(unitLengthField.data(), unitLengthField.size());
+	}	
+
+private:
+	
+	std::vector<uint8_t> unitLengthField;
+	
+};
+
+class H264FrameMediaAppender : public H26xFrameMediaAppender
```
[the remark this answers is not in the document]
harryz2000: Yep. Good point. Was also planing to do so.

include/avcdescriptor.h
```diff
@@ -51,6 +49,37 @@ class AVCDescriptor
 	void SetNALUnitLengthSizeMinus1(BYTE NALUnitLengthSizeMinus1)			{ this->NALUnitLengthSizeMinus1 = NALUnitLengthSizeMinus1;			}
 
 
+	virtual bool GetResolution(unsigned& width, unsigned& height) const
+	{
+		for (int i=0;i<GetNumOfSequenceParameterSets();i++)
```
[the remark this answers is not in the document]
harryz2000: Yep. Not sure yet. Need to double check. The decoder might be better to tell whether the width and height has been got. This code is basically try to follow the existing code (although changed to return immediately).

include/H26xNaluPacketizer.h
harryz2000: I would prefer to keep it separately. Adding to one file sometimes make the file has too many dependencies IMO. We would have more for other codecs, such as Av1OBUPacketizer, e.g.

src/h264/H264NaluPacketizer.h
harryz2000: I would prefer to keep it separately. Adding to one file sometimes make the file has too many dependencies IMO. We would have more for other codecs, such as Av1OBUPacketizer, e.g.

src/h265/H265NaluPacketizer.h
harryz2000: Yes. I also prefer Constants.

include/avcdescriptor.h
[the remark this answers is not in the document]
harryz2000: This should be safe as the width/height are compulsory in every SPS. The first one would be enough. What do you mean "repetitive" info?

PR #182, opened 2023-07-28
Fix crash caused by RTMPConnection shared_from_this

harryz2000:


src/rtmp/rtmpconnection.cpp
```diff
@@ -102,11 +102,11 @@ void RTMPConnection::Start()
 	running = true;
 	
 	//Start thread and run
-	thread = std::thread([=](){
+	thread = std::thread([=, connection=shared_from_this()](){
 		//Block signals to avoid exiting on SIGUSR1
 		blocksignals();
 		//Run
-		Run(); 
+		Run(connection);
```
murillo128, line 109: no need to pass the connection to the Run command, the `connection` shared_ptr should be alive during the whole life of the lamda execution.

src/rtmp/rtmpconnection.cpp
```diff
@@ -102,11 +102,11 @@ void RTMPConnection::Start()
 	running = true;
 	
 	//Start thread and run
-	thread = std::thread([=](){
+	thread = std::thread([=, connection=shared_from_this()](){
```
murillo128, line 105: rename to `self` (that's what we use in other parts of the code similar to this)

murillo128 (changes requested):
How do you think the segmentation is happening?

The only way I can see it from happening is if `End()` is called before the thread is started and it dosen't wait for the thread to end.  But in this case all the Run() instructions would be using garbage mem for this.

src/rtmp/rtmpconnection.cpp
harryz2000: The issue is if the thread is executed during destructor, the shared_ptr is not available.

harryz2000: This is an example that would throw:
```
#include <memory>

class Test : public std::enable_shared_from_this<Test>
{
public:
    ~Test()
    {
        auto a = shared_from_this();
    }
};


int main()
{
    auto b = std::make_shared<Test>();
    return 0;
}
```

src/rtmp/rtmpconnection.cpp
harryz2000: Done.

src/rtmp/rtmpconnection.cpp
murillo128: It makes sense, good catch!

src/rtmp/rtmpconnection.cpp
```diff
@@ -102,11 +102,11 @@ void RTMPConnection::Start()
 	running = true;
 	
 	//Start thread and run
-	thread = std::thread([=](){
+	thread = std::thread([=, self=shared_from_this()](){
 		//Block signals to avoid exiting on SIGUSR1
 		blocksignals();
 		//Run
-		Run(); 
+		Run(self);
```
murillo128, line 109: I don't think we need to pass `self` to the `Run()` method, as it is being captured in the lambda, also should we try to get a weak reference instead to prevent leaking?

murillo128: like 

```
thread = std::thread([=, weak=std::weak_ptr<RTMPConnection>(shared_from_this())](){
                //Check if the shared pointer is still valid
                auto self = weak.lock();
                //If connection has already ended
                if (!self)
                      //Exit
                      return;
		//Block signals to avoid exiting on SIGUSR1
		blocksignals();
		//Run
		Run();
          });
````

Comment (murillo128) [reply src/rtmp/rtmpconnection.cpp:109]: Although in this case we wouldn't be calling the `onDisconnect` callback, so probably just better to do

```
//Start thread and run, hold reference to us to prevent being destroyed before Run ends.
thread = std::thread([=, self=shared_from_this()](){
		//Block signals to avoid exiting on SIGUSR1
		blocksignals();
		//Run
		Run();
          });
```

murillo128 (changes requested):

src/rtmp/rtmpconnection.cpp
harryz2000: > like
> 
> ```
> thread = std::thread([=, weak=std::weak_ptr<RTMPConnection>(shared_from_this())](){
>                 //Check if the shared pointer is still valid
>                 auto self = weak.lock();
>                 //If connection has already ended
>                 if (!self)
>                       //Exit
>                       return;
> 		//Block signals to avoid exiting on SIGUSR1
> 		blocksignals();
> 		//Run
> 		Run();
>           });
> ```

Yes. This would skip the onDisconnect, which should not what we want.

harryz2000: > ```
> //Start thread and run, hold reference to us to prevent being destroyed before Run ends.
> thread = std::thread([=, self=shared_from_this()](){
> 		//Block signals to avoid exiting on SIGUSR1
> 		blocksignals();
> 		//Run
> 		Run();
>           });
> ```

I think this one works same as passing the the shared_ptr to Run() function. I am not sure which is better. Passing it as argument looks more explicit to me. But I am fine to change to this.

harryz2000: Done.

harryz2000: By the way, about the leaking concern. The lambda would be destroyed once the thread is stopped, so is the shared_ptr. And I tested the RTMPConnection could be destroyed. The only thing would be the End() call in the destructor wouldn't be too much useful as the thread must be stopped so it could be released.

murillo128 (approved):

PR #183, opened 2023-07-31
Ensure timer callback wouldn't be called after destruction

harryz2000:


harryz2000:
I think it would be ideal the TimerService (EventLoop) could provide a "synced" "cancel" function to have something similar to set "timerCancelled" flag so after the cancel, the callback wouldn't be executed.

src/MediaFrameListenerBridge.cpp
```diff
@@ -73,6 +75,12 @@ MediaFrameListenerBridge::MediaFrameListenerBridge(TimeService& timeService,DWOR
 MediaFrameListenerBridge::~MediaFrameListenerBridge()
 {
 	Debug("-MediaFrameListenerBridge::~MediaFrameListenerBridge() [this:%p]\n", this);
+	
+	dispatchTimer->Cancel();
+	
+	timeService.Sync([this](auto){
+		timerCancelled = true;
```
murillo128, line 82: You can do this instead:

```
timeService.Sync([this](auto){
    dispatchTimer->Cancel();
})
```

Given that the both the timer lamba and the dispatch timer cancelation will be run on the EventLoop thread, there is no need for the `timerCancelled` flag. 

Also, better move this to the `Stop()` call to prevent a race condition between the `MediaFrameListener` destructor and the `EventLoop` destructor (being js garbage collected objects we can't ensure the disposal order)

murillo128 (changes requested):

src/MediaFrameListenerBridge.cpp
murillo128: something like

```
void MediaFrameListenerBridge::Stop()
{
	Debug("-MediaFrameListenerBridge::Stop() [this:%p]\n", this);

	timeService.Sync([=](auto now) {
                //Dispatch pendingREADY packets now
		std::vector<RTPPacket::shared> pending;		
		while (packets.size())
		{
			//Add to pending packets
			pending.push_back(packetInfo.packet);
			//remove it from pending packets
			packets.pop();
		}
		//Dispatch RTP packets
		Dispatch(pending);

                //Updated last dispatched 
		lastSent = now;

		//Cancel dispatch timer
                dispatchTimer->Cancel();

		//TODO wait onMediaFrame end
		for (auto listener : listeners)
			listener->onEnded(this);
		//Clear listeners
		listeners.clear();

                //Flag for not schedulin the dispatch timer again
                stopped = true;                
	});
}
```

we should also check for the `stopped` flag just before:
https://github.com/medooze/media-server/blob/34a56ba708735e1bc1538e453fd7a7a44a4154c2/src/MediaFrameListenerBridge.cpp#L149

harryz2000: Done as suggested. 

Still think having a function like `CancelRemoveCallbacks() `would be handy. But not a priority.

src/MediaFrameListenerBridge.cpp
```diff
@@ -73,18 +75,45 @@ MediaFrameListenerBridge::MediaFrameListenerBridge(TimeService& timeService,DWOR
 MediaFrameListenerBridge::~MediaFrameListenerBridge()
 {
 	Debug("-MediaFrameListenerBridge::~MediaFrameListenerBridge() [this:%p]\n", this);
+	
+	Stop();
```
murillo128, line 79: don't call Stop()  in the destructor, there is no warranty that the EventLoop will be alive at this time (because of js garbage collection)

murillo128 (changes requested):

src/MediaFrameListenerBridge.cpp
harryz2000: If that is case, should we pass a shared_ptr<TimeService> instead? Or keep the TimeService live during MediaFrameListenerBridge? This uncertainty seems not ideal.

Plus, if MediaFrameListenerBridge is released but TimeService is not, would the stop() call be guaranteed by other code before the release? Otherwise, it might still have the crash issue.

harryz2000: If the Stop() would be called before destruction, this stop() call wouldn't cause any issue as it would return immediately.

murillo128 (approved):

src/MediaFrameListenerBridge.cpp
murillo128: i didn't saw the check in the `Stop()` method, lgtm now.

PR #186, opened 2023-08-14
RTMPConnection: expose peer address

mildsunrise:
when reviewing code I found a TODO for this, thought it was quick enough to implement

### DIFF
--- a/include/rtmp/rtmpapplication.h
+++ b/include/rtmp/rtmpapplication.h
@@ -4,11 +4,17 @@
 #include "rtmpnetconnection.h"
 #include <string>
 #include <functional>
+#include <sys/socket.h>
 
 class RTMPApplication
 {
 public:
-	virtual RTMPNetConnection::shared Connect(const std::wstring& appName,RTMPNetConnection::Listener *listener,std::function<void(bool)> accept) = 0;
+	virtual RTMPNetConnection::shared Connect(
+		const struct sockaddr_in& peername,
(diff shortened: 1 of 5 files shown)


include/rtmp/rtmpapplication.h
```diff
@@ -4,11 +4,17 @@
...
 #include <string>
 #include <functional>
+#include <sys/socket.h>
 
 class RTMPApplication
 {
 public:
-	virtual RTMPNetConnection::shared Connect(const std::wstring& appName,RTMPNetConnection::Listener *listener,std::function<void(bool)> accept) = 0;
+	virtual RTMPNetConnection::shared Connect(
+		struct sockaddr_in peername,
```
murillo128, line 13: do we need the `struct`? pass by const & better

src/rtmp/rtmpconnection.cpp
```diff
@@ -945,6 +945,16 @@ void RTMPConnection::ProcessCommandMessage(DWORD streamId,RTMPCommandMessage* cm
 			//Send error
 			return SendCommandError(streamId,transId);
 
+		//Get client address
+		struct sockaddr_in peername;
+		socklen_t peername_len = sizeof(peername);
```
murillo128, line 950: use cammel case

murillo128 (commented):
good one, just minor nits

mildsunrise:
both things fixed, merging

PR #198, opened 2023-09-05
Static analysis: AMFNumber parsing not handle negative value

harryz2000:


src/rtmp/amf.cpp
```diff
@@ -463,7 +463,7 @@ double AMFNumber::GetNumber()
 {
 	if(value+value > 0xFFEULL<<52)
         	return 0.0;
-    	return ldexp(((value&((1LL<<52)-1)) + (1LL<<52)) * (value>>63|1), (value>>52&0x7FF)-1075);
+    	return std::ldexp(((value&((1LL<<52)-1)) + (1LL<<52)), (value>>52&0x7FF)-1075) * (((value>>63) == 1) ? -1 : 1);
```
murillo128, line 466: are we sure about this? cc/ @mildsunrise ?

harryz2000: Yes. Added unit tests

murillo128: were did you got the bytes/number conversion from?

harryz2000: The AMF number is defined in https://rtmp.veriskope.com/pdf/amf0-file-format-specification.pdf. It basically says it is a "8 byte IEEE-754 double floating point value".

The IEEE-754 double format can be found at: https://en.wikipedia.org/wiki/Double-precision_floating-point_format

In the unit tests, I created some random values for testing, including the limits. Some are from the WIKI page, some I generated through https://www.binaryconvert.com/convert_double.html

PR #199, opened 2023-09-05
Static analysis: various defects

harryz2000:
Potential nullptr use, divide by zero, incorrect count calculation, always false condition, etc


mildsunrise:
the two diffs under `src/rtp` are huge, could it be because of line endings?

harryz2000:
Yeah. I also found it. Looks like a merge issue. I am fixing it.

harryz2000:
Yes. It turns out the line ending of these files are CRLF (different with others) in the repository. I manually changed these two files to be CRLF and it appears working now. 

We should unify the line endings one day IMO.

src/aac/aacconfig.h
```diff
@@ -74,7 +74,7 @@ class AACSpecificConfig
 	static BYTE GetSampleRateIndex(DWORD rate)
 	{
 		//Search
-		for (DWORD i=0;i<sizeof(rates);i++)
+		for (DWORD i=0;i<sizeof(rates)/sizeof(rates[0]);i++)
```
murillo128, line 77: should we change `rates` to an std::array better?

harryz2000: I am OK for both. But change to std::array, which seems more modern.

src/aac/aacconfig.h
```diff
@@ -58,7 +58,7 @@
 class AACSpecificConfig
 {
 public:
-	static constexpr DWORD rates[13] = {
+	static constexpr std::array<DWORD, 13> rates = {
```
murillo128, line 61: I don't think that the `13` is needed

murillo128 (changes requested):

src/aac/aacconfig.h
harryz2000: The std::array needs two template arguments. Using one doesn't compile.

murillo128 (approved):

PR #200, opened 2023-09-05
Static analysis: trivial format specifier, return value check, etc

harryz2000:


src/DTLSICETransport.cpp
```diff
@@ -2245,8 +2246,7 @@ int DTLSICETransport::Send(const RTPPacket::shared& packet)
 		//Append it to the end of the packet history
 		history.push_back(packet);
 	
-	//Did we send successfully?
-	return (len>=0);
```
harryz2000, line 2249: Note, len is always >= 0 here.

src/DTLSICETransport.cpp
```diff
@@ -463,7 +464,7 @@ int DTLSICETransport::onData(const ICERemoteCandidate* candidate,const BYTE* dat
 	
 	//Send nack feedback
 	if ((group->type == MediaFrame::Video || recvMaps.apt.GetTypeForCodec(packet->GetPayloadType()!=RTPMap::NotFound)) &&
-		( lost>0 ||  (group->GetCurrentLost() && (now-source->lastNACKed)/1000>fmax(rtt,20)))
+		( lost>0 ||  (group->GetCurrentLost() && (now-source->lastNACKed)/1000>std::max(rtt,uint32_t(20))))
```
murillo128, line 467: i prefer `std::max<uint32_t>(rtt,20);` or `std::max<rtt,20u>;` not sure which one of the 3 is more correct

src/h264/H264LayerSelector.cpp
```diff
@@ -4,6 +4,7 @@
 H264LayerSelector::H264LayerSelector()
 {
 	waitingForIntra = true;
+	nextTemporalLayerId = 0;
```
murillo128, line 7: delete from constructor as they are in .h

src/avcdescriptor.cpp
```diff
@@ -262,7 +263,7 @@ void AVCDescriptor::Dump() const
 	{
 		H264SeqParameterSet sps;
 		//Decode
-		sps.Decode(spsData[i],spsSizes[i]);
+		(void)sps.Decode(spsData[i],spsSizes[i]);
```
murillo128, line 266: change by `if (sps.Decode(...`

murillo128 (changes requested):

include/rtmp/rtmpclientconnection.h
```diff
@@ -151,8 +151,8 @@ class RTMPClientConnection :
 			this->tag = tag;
 		}
 		
-		TransType type;
-		DWORD par;
+		TransType type {};
```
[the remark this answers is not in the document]
harryz2000: Tried to give it a zeroed value since as long as it has a certain value it wouldn't cause strange issues. Would change to an valid value, which probably better for readibility.

include/rtp/RTCPPayloadFeedback.h
```diff
@@ -447,9 +452,9 @@ class RTCPPayloadFeedback : public RTCPPacket
 private:
 	typedef std::vector<Field::shared> Fields;
 private:
-	FeedbackType feedbackType;
-	DWORD senderSSRC;
-	DWORD mediaSSRC;
+	FeedbackType feedbackType {};
```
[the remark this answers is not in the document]
harryz2000: As above. Will change.

src/avcdescriptor.cpp
harryz2000: Yeah. Agree. Considered that but kept no change for keeping the change trivial. Would fix then.

include/rtmp/rtmpclientconnection.h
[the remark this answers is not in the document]
murillo128: We should probably either remove any constructor that doesn't set the TransType or add an Unknown TransType

include/CPUMonitor.h
```diff
@@ -34,10 +34,10 @@ class CPUMonitor
         static void * run(void *par);
 
 private:
-	int		numcpu;
-	int		interval;
-	bool		started;
-	pthread_t	thread;
+	int		numcpu = 0;
+	int		interval = 0;
+	bool		started = false;
```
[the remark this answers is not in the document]
harryz2000: Done

include/rtmp/rtmpclientconnection.h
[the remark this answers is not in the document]
harryz2000: Removed the default constructor, which seems not very useful

src/h264/H264LayerSelector.cpp
harryz2000: OK

src/DTLSICETransport.cpp
harryz2000: both looks OK. Preferred 20u one as it is shorter.

murillo128 (approved):

PR #205, opened 2023-09-07
Add dispatch delay limit

harryz2000:
Limit the delay to a max value to avoid invalid/illegal packets causing big delay, therefore causing memory issue.

### DIFF
--- a/src/MediaFrameListenerBridge.cpp
+++ b/src/MediaFrameListenerBridge.cpp
@@ -119,7 +119,7 @@ void MediaFrameListenerBridge::Stop()
 
 void MediaFrameListenerBridge::SetDelayMs(std::chrono::milliseconds delayMs)
 {
-	timeService.Async([&](auto now) { 
+	timeService.Async([=](auto now) { 
 		// Set a delay limit so the queue wouldn't grow without control if something is wrong, i.e. the packet
 		// time info was not valid
 		constexpr int64_t MaxDelayMs = 5000;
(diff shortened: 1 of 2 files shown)


src/MediaFrameListenerBridge.cpp
```diff
@@ -120,7 +120,18 @@ void MediaFrameListenerBridge::Stop()
 void MediaFrameListenerBridge::SetDelayMs(std::chrono::milliseconds delayMs)
 {
 	timeService.Async([&](auto now) {
```
murillo128, line 122: should this be a
`
timeService.Async([=](auto now) {
`
instead?

test/unit/TestCommon.h
```diff
@@ -25,13 +25,17 @@ class TestTimeService : public TimeService
 	
 		virtual void Cancel() override
 		{
-			assert(false); // Not used
+			for (auto& schedule : loop.pendingTimers)
```
murillo128, line 28: lgtm but please, move to a different PR

murillo128 (approved):

src/MediaFrameListenerBridge.cpp
harryz2000: Yes. Indeed.

test/unit/TestCommon.h
harryz2000: Done

murillo128 (approved):

PR #212, opened 2023-09-21
Add support for multi-payload-unit h264 frames

harryz2000:


src/h264/h264.h
```diff
@@ -285,6 +285,62 @@ class H264PictureParameterSet
 	bool			redundant_pic_cnt_present_flag = false;
 };
 
+class H264SliceHeader
+{
+public:
+	H264SliceHeader(const H264SeqParameterSet& sps) : sps(sps)
```
murillo128, line 291: pass the SPS as param to the Decode instead

src/h264/h264.h
```diff
@@ -285,6 +285,62 @@ class H264PictureParameterSet
...
+	{
+	}
+
+	bool Decode(const BYTE* buffer,DWORD bufferSize)
+	{	
+		BitReader r(buffer, bufferSize);
+		
+		[[maybe_unused]] auto firstMbInSlice = ExpGolombDecoder::Decode(r); CHECK(r);
```
murillo128, line 299: Add `firstMBInSlice` and the other decoded properties as attributes to the class and add getters for it

src/h264/h264.h
```diff
@@ -285,6 +285,62 @@ class H264PictureParameterSet
...
+	}
+	
+	inline bool IsFrameEnd() const
+	{
+		return frameEnd;
+	}
+	
+private:
+	const H264SeqParameterSet& sps;
+	bool frameEnd = false;
```
murillo128, line 341: We may want to reuse this class for other stuff  so I would prefer this class as data only and use the spec name `bottomFieldFlag` for the attribute instead of `frameEnd`.

murillo128 (changes requested):
lgtm, just some minor editorial changes

src/h264/H26xPacketizer.h
```diff
@@ -19,9 +19,11 @@ class H26xPacketizer : public RTPPacketizer
 	H26xPacketizer(VideoCodec::Type codec);
 	virtual std::unique_ptr<MediaFrame> ProcessAU(BufferReader& payload);
 
+	void resetFrame();
```
murillo128, line 22: Make the first `r` upper case

src/h264/h264.h
harryz2000: Done

src/h264/h264.h
harryz2000: Done.

src/h264/H26xPacketizer.h
harryz2000: Done

murillo128 (approved):
lgtm

PR #214, opened 2023-09-26
[W.I.P] Improve CMake and enable end-of-stream/file VP8&H264 encoding 

ZitaLiao:
This are mainly to support my4 file transcoder
1. make MediaServerLib a more self-contain lib.
   a. Improve Cmake
   b. import medooze/mp4v2 as submodule
2. make VP8 & H264 support end-of-stream/file encoding

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "ext/googletest"]
 	path = ext/googletest
 	url = https://github.com/google/googletest.git
+[submodule "ext/mp4v2"]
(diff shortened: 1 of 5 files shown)


ZitaLiao:
@murillo128 I've changed mp4v2 from submodule to Cmake FetchContent. Could you have a look again?
It would be better if you could merge https://github.com/medooze/mp4v2/pull/1 first then we can point to the mp4v2 main branch here as well.

murillo128 (approved):
lgtm

PR #217, opened 2023-10-06
Fixed static analysis errors

harryz2000:


src/h265/H265Packetizer.cpp
```diff
@@ -164,7 +164,7 @@ void H265Packetizer::OnNal(VideoFrame& frame, BufferReader& reader, std::optiona
 	// just before the first slice of the frame:
 	if (
 		// this NALU is a slice or SEI
-		((nalUnitType >= HEVC_RTP_NALU_Type::TRAIL_N && nalUnitType <= HEVC_RTP_NALU_Type::RASL_R)
+		((nalUnitType <= HEVC_RTP_NALU_Type::RASL_R)
```
murillo128, line 167: @ZitaLiao can you double check this?

src/SimulcastMediaFrameListener.cpp
```diff
@@ -277,6 +279,8 @@ void SimulcastMediaFrameListener::Flush()
 	{
 		auto f = std::move(queue.front());
 		queue.pop_front();
+// Ignore coverity error: Attempting to access the managed object of an empty smart pointer "f".
+// coverity[dereference]
```
murillo128, line 283: Do we know why coverity is complaining about this?

harryz2000: The detailed report can be found in the artifact on result https://github.com/medooze/static-analysis/actions/runs/6429275751.
It looks the Coverity is confused about the effect of pop_front(), which has removed the empty front. Coverity thinks we are still using the front(). It appears a false alarm.

src/h265/H265Packetizer.cpp
ZitaLiao: Fine here since HEVC_RTP_NALU_Type::TRAIL_N is 0. 
@harryz2000 A better way may be to keep it with a commented line to indicate clearer information though, minor improvement.

harryz2000: That is exactly what the Coverity complain about. The nalUnitType is an unsigned type and TRAIL_N is 0. The condition would be never false.

harryz2000: I assume TRAIL_N value  (as well as other values with in the range) would never change. Otherwise, we need to take care of every value in between.

src/SimulcastMediaFrameListener.cpp
harryz2000: Plus, I tested similar code logic and it didn't have issues.

murillo128 (approved):
lgtm

PR #218, opened 2023-10-06
Add pixel aspect ratio

murillo128:


include/VideoBuffer.h
```diff
@@ -101,6 +101,11 @@ class VideoBuffer
 	ColorRange GetColorRange() const { return colorRange; }
 	ColorSpace GetColorSpace() const { return colorSpace; }
 
+	void SetPixelAspectRatio(int pixelAspectRatioNum, int pixelAspectRatioDen)	{ this->pixelAspectRatio = { pixelAspectRatioNum , pixelAspectRatioDen }; }
+	void SetPixelAspectRatio(std::pair<int,int> pixelAspectRatio)			{ this->pixelAspectRatio = pixelAspectRatio; }
+	std::pair<int, int> GetPixelAspectRatio() const			{ return this->pixelAspectRatio; }
+	bool HasNonSquarePixelAspectRatio() const			{ return this->pixelAspectRatio.first && this->pixelAspectRatio.first != this->pixelAspectRatio.second; }
```
ZitaLiao, line 107: Why do we need to check this->pixelAspecRatio.first   here? To check it not 0?

src/VideoPipe.cpp
```diff
@@ -154,6 +154,30 @@ VideoBuffer::const_shared VideoPipe::GrabFrame(uint32_t timeout)
 	//Unlock
 	pthread_mutex_unlock(&newPicMutex);
 
+	//If we have a dinamic resize
+	if (scaleResolutionDownBy > 0 || videoBuffer->HasNonSquarePixelAspectRatio())
```
ZitaLiao, line 158: minor: check if videoBuffer is not null before using the pointer?

src/VideoPipe.cpp
```diff
@@ -154,6 +154,30 @@ VideoBuffer::const_shared VideoPipe::GrabFrame(uint32_t timeout)
...
+	{
+		//Get pixel aspect ratio
+		const auto [parNum, parDen] = videoBuffer->GetPixelAspectRatio();
+
+		//Choose which dimension we want to use for adjusting the pixel aspect ratio
+		if (parNum>parDen)
+		{
+			//Check adjusted video size
+			videoWidth = ((uint32_t)((videoBuffer->GetWidth() * parNum) / (scaleResolutionDownBy * parDen))) & ~1;
```
ZitaLiao, line 167: need to check if scaleResolutionDownBy is 0, or else would crash by devided by 0. Same in other places

ZitaLiao (approved):
lgtm except the minor null/0-den check on the comment, they would not fail with current usage/code though.

PR #220, opened 2023-10-09
Fix static analysis errors relevant to video codecs

harryz2000:


src/h264/h264encoder.cpp
```diff
@@ -68,6 +68,15 @@ H264Encoder::H264Encoder(const Properties& properties) : frame(VideoCodec::H264)
 	type    = VideoCodec::H264;
 	format  = 0;
 	pts     = 0;
+	curNal = 0;
```
murillo128, line 71: can we move it to the .h instead?

src/vp8/vp8encoder.cpp
```diff
@@ -31,7 +31,16 @@ VP8Encoder::VP8Encoder(const Properties& properties) : frame(VideoCodec::VP8)
 	type    = VideoCodec::VP8;
 	pts	= 0;
 	num	= 0;
-	pic	= NULL;
+	pic	= NULL;
```
murillo128, line 34: can we move it to the .h instead?

src/h264/h264encoder.cpp
```diff
@@ -68,6 +68,15 @@ H264Encoder::H264Encoder(const Properties& properties) : frame(VideoCodec::H264)
...
+	numNals = 0;
+	width = 0;
+	height = 0;
+	nals = 0;
+	
+	memset(&pic, 0, sizeof(pic));
```
murillo128, line 77: `pic = {}` in .h instead?

include/VideoDecoderWorker.h
```diff
@@ -32,7 +32,7 @@ class VideoDecoderWorker
 private:
 	std::set<VideoOutput*> outputs;
 	WaitQueue<RTPPacket::shared> packets;
-	pthread_t thread;
+	pthread_t thread = 0;
```
murillo128, line 35: nullptr or {} better?

murillo128 (changes requested):

src/vp8/vp8encoder.cpp
harryz2000: I tried to make small change and keep the initialisation consistent with existing code. Would you like to change the existing code as well? Sometimes the change would be big and probably not be convenient for review.

include/VideoDecoderWorker.h
harryz2000: As mentioned in another pr. pthread_t is an integer, seems 0 is more suitable

src/h264/h264encoder.cpp
harryz2000: Yes. We can. As mentioned in another response, how should we resolve this as existing code does initialisation in cpp file. Which way should we take?
1. moving all initialisation to header file, including existing code
2. Only add initialisation for missing ones, don't change existing ones
3. Keep the missing ones consistent with existing code
I am fine to either way.

murillo128: The code may be bigger, but the complexity is low. So let's modernize it and go with 3.

harryz2000: I assume you meant: Make all initialisation consistent and in header file. I will change to that way. Please let me know if my understanding is correct.

murillo128: that's correct

src/h264/h264encoder.cpp
harryz2000: Done

src/vp8/vp8encoder.cpp
harryz2000: Done

src/h264/h264encoder.cpp
harryz2000: Done.

harryz2000: Actually let me double check whether {} has same effect as memset. The {} does eliminate the Coverity errors.

harryz2000: Yes. My tests indicated it does zero the struct.

murillo128 (approved):
lgtm

PR #221, opened 2023-10-09
Fix audio codec relevant static analysis errors

harryz2000:


include/audiodecoder.h
```diff
@@ -40,7 +40,7 @@ class AudioDecoderWorker
 private:
 	std::set<AudioOutput*> outputs;
 	WaitQueue<RTPPacket::shared> packets;
-	pthread_t thread;
+	pthread_t thread = 0;
```
murillo128, line 43: wouldn't nullptr or {} be better than 0?

src/AudioPipe.cpp
```diff
@@ -173,15 +173,18 @@ int AudioPipe::PlayBuffer(SWORD* buffer, DWORD size, DWORD frameTime, BYTE vadLe
 	}
 
 	//Check if we are transtrating
+	SWORD resampled[8192];
```
murillo128, line 176: why is this needed?

src/audioencoder.cpp
```diff
@@ -29,6 +29,9 @@ AudioEncoderWorker::AudioEncoderWorker()
 	audioCodec=AudioCodec::PCMU;
 	//Create mutex
 	pthread_mutex_init(&mutex,0);
+	
+	audioInput = nullptr;
```
murillo128, line 33: move to audioencoder.h

src/opus/opusencoder.cpp
```diff
@@ -55,6 +59,8 @@ DWORD OpusEncoder::TrySetRate(DWORD rate, DWORD numChannels)
 		numFrameSamples = rate * 20 / 1000;
 	}
 
+	if (enc == nullptr) return 0;
```
murillo128, line 62: use `!enc` better to be consistent with the other checks in the file

src/opus/opusencoder.cpp
```diff
@@ -70,5 +76,7 @@ OpusEncoder::~OpusEncoder()
 
 int OpusEncoder::Encode(SWORD *in,int inLen,BYTE* out,int outLen)
 {
+	if (enc == nullptr) return OPUS_BAD_ARG;
```
murillo128, line 79: `return 0;` better instead

include/audiodecoder.h
harryz2000: pthread_t is actually a 'unsigned long int'. Using 0 seems more suitable.

src/opus/opusencoder.cpp
harryz2000: 0 actually is same as OPUS_OK. I used OPUS_BAD_ARG to make it be compatible with opus_encode. Does that make sense?

src/AudioPipe.cpp
harryz2000: because resampled is assigned to buffer, which is used in the function scope. Previously, resampled is declared in the if condition scope.

src/opus/opusencoder.cpp
murillo128: we use -1 or 0 to indicate an error in encoding:

https://github.com/medooze/media-server/blob/master/src/audioencoder.cpp#L238C1-L246C1

harryz2000: OPUS_BAD_ARG is actually -1. I can change it to -1 if you prefer.
```
[00051](https://www.opus-codec.org/docs/html_api/group__errorcodes.html#gaa44cf8a185e1b5cb940ef63eb4f02d21) #define OPUS_OK                0
00052 
[00053](https://www.opus-codec.org/docs/html_api/group__errorcodes.html#gaf2d43e479455a1a3b6874e5faf4e827d) #define OPUS_BAD_ARG          -1
00054 
[00055](https://www.opus-codec.org/docs/html_api/group__errorcodes.html#gacd897c05563ec04a67d8e92ba29f3d4f) #define OPUS_BUFFER_TOO_SMALL -2
00056
```

src/opus/opusencoder.cpp
harryz2000: OK

src/audioencoder.cpp
harryz2000: OK

harryz2000: I moved all the variable initialisation to header file. Otherwise, some in cpp while others in header looks strange to me.

src/opus/opusencoder.cpp
harryz2000: Use -1 now

murillo128 (approved):
lgtm

PR #223, opened 2023-10-13
Coverity fix: pointer/resource leak

ZitaLiao:


murillo128:
@harryz2000 could we try to use `unique_ptr` to manage the leaked resources instead?

harryz2000:
Yes. Using unique_ptr would be a good idea. Will change.

harryz2000:
Yes. That would be a good idea. Will make the change.

src/mp4streamer.cpp
```diff
@@ -996,7 +996,7 @@ AVCDescriptor* MP4Streamer::GetAVCDescriptor()
 	if (pictureHeaderSize)
 		free(pictureHeaderSize);
 
-	return desc;
+	return desc.release();
```
murillo128, line 999: let's make `MP4Streamer::GetAVCDescriptor()` return a `std::unique_ptr` and chante the rtmpmp4streamer to use a unique ptr too

src/mp4recorder.cpp
```diff
@@ -912,12 +911,13 @@ void MP4Recorder::processMediaFrame(DWORD ssrc, const MediaFrame &frame, QWORD t
...
-					videoTrack = new mp4track(mp4);
+					auto videoTrackPtr = std::make_unique<mp4track>(mp4);
 					//Create track
-					videoTrack->CreateVideoTrack(videoFrame.GetCodec(),videoFrame.GetClockRate(),videoFrame.GetWidth(),videoFrame.GetHeight(),disableHints);
+					videoTrackPtr->CreateVideoTrack(videoFrame.GetCodec(),videoFrame.GetClockRate(),videoFrame.GetWidth(),videoFrame.GetHeight(),disableHints);
 					//Set name as ssrc
-					videoTrack->SetTrackName(std::to_string(ssrc));
+					videoTrackPtr->SetTrackName(std::to_string(ssrc));
 					//Add it to map
+					videoTrack = videoTrackPtr.release();
```
murillo128, line 920: I think that we should modernize the whole pointer management of the tracks. So let's change the `videoTracks` to be a vector of `shared_ptr`s and use shared_ptrs instead of unique_ptrs for `videoTrack` so the code here is easier to change

murillo128 (changes requested):
soory for changin my mind, but let's try to modernize the whole file

src/rtmp/rtmpmp4stream.cpp
```diff
@@ -130,7 +130,7 @@ void RTMPMP4Stream::doPlay(QWORD transId, std::wstring& url,RTMPMediaStream::Lis
 	SendMetaData(meta);
 
 	//Get AVC descriptor if any
-	desc = streamer.GetAVCDescriptor();
+	desc = streamer.GetAVCDescriptor().release();
```
murillo128, line 133: can we make `desc`  in in RTMPMP4Stream a unique ptr too?

harryz2000: It seems this rtmpmp4stream.cpp wasn't used in any build files. Therefore, I tented to do minor changes in case the change would break anything. Basically I didn't optimize this file. Where was this file used? Or should we optimize it? If so, we might want to add it to some cmake file to at least verify it can compile.

murillo128: ok, let's ignore it for now.

murillo128 (approved):
lgtm

PR #224, opened 2023-10-24
DIOS-2757 Remove init of rand num. Fix crash on native thread

lstac:


src/RTPBundleTransport.cpp
```diff
@@ -103,6 +103,13 @@ RTPBundleTransport::RTPBundleTransport(uint32_t packetPoolSize) :
 	//Init values
 	socket = FD_INVALID;
 	port = 0;
+
+	//Init random
+	static init_seed = false;
```
murillo128, line 108: I think we can do this at the the media server initialization instead

murillo128 (changes requested):
lgtm except for the seed initialization

mildsunrise (approved):

PR #225, opened 2023-11-14
Add bframes, iframes stats to publishing feed measurement

harryz2000:
Also fixed frame resolution is not always set for RTMP frames.
Note since total frames, bframes and iframes are all will be in stats. The pframe information then can be deduced from the data.


murillo128 (approved):
lgtm

PR #228, opened 2023-11-23
Set RTT to streams. RTT is only checked once for now.

harryz2000:


src/rtmp/rtmpconnection.cpp
```diff
@@ -844,12 +844,21 @@ void RTMPConnection::ProcessControlMessage(DWORD streamId,BYTE type,RTMPObject*
...
 					break;
 				case RTMPUserControlMessage::PingResponse:
+				{
 					//Get response
 					DWORD ping = event->GetEventData();
 					//Get roundtrip delay
 					rtt = getDifTime(&startTime)/1000-ping;
+					
+					//Check if a stream has been created with that id
+					pthread_mutex_lock(&mutex);
```
murillo128, line 854: can we use range for loops instead?

murillo128 (changes requested):
minor change required, lgtm otherwise

src/rtmp/rtmpconnection.cpp
harryz2000: Yes. Done

murillo128 (approved):
lgtm

PR #230, opened 2023-11-23
DIOS-2863 Add codecs in use to getstatsdb

bcostdolby:


include/rtp/RTPIncomingSourceGroup.h
```diff
@@ -73,6 +73,7 @@ class RTPIncomingSourceGroup :
 	long double avgWaitedTime = 0;
 	
 	QWORD lastUpdated = 0;
+	DWORD codec = UINT32_MAX;
```
murillo128, line 76: `codec` is a `BYTE` in `RTPPacket`, I think we should keep same type and make it optional
```suggestion
	std::optional<BYTE> codec;
```

bcostdolby: I ended up just going with a BYTE instead of optional<BYTE> as then UNKNOWN will be handled in exactly the same was as everything else does (by truncating it to 0xFF) and the code was a bit tidier. At some point it would bee good to review the types used for codecs but as you said now is not it.

murillo128 (approved):
lgtm

PR #231, opened 2023-11-28
fix STUN keepalives

mildsunrise:
otherwise, the presence of connections that send keepalives but never receive an answer back causes onTimer() to never go to the second half where it sends keepalives for the healthy connections.

the effect is that the RTPBundleTransport is kept busy sending keepalives to a select number of connections, while keepalives are effectively disabled for the rest.


murillo128 (approved):
lgtm

PR #232, opened 2023-11-30
Add timestamp checker

harryz2000:


include/TimeStampChecker.h
```diff
@@ -0,0 +1,71 @@
+#ifndef TIMESTAMPCHECKER_H
+#define TIMESTAMPCHECKER_H
+
+#include <media.h>
+#include <acumulator.h>
+
+class TimeStampChecker
+{
+public:
+	static constexpr uint32_t MaxDurationDiffMs = 3000;
```
murillo128, line 10: this should be a configurable parameter on the constructor of `TimeStampChecker`

murillo128: Also, please add a brief description of what does this class do

murillo128 (changes requested):

include/TimeStampChecker.h
```diff
@@ -0,0 +1,71 @@
...
+	static constexpr uint32_t MaxContinousInvalidFrames = 5;
+	
+	enum class CheckResult
+	{
+		Valid,
+		Invalid,
+		Reset
+	};
+	
+	CheckResult Check(const MediaFrame &frame)
```
murillo128, line 20: If it is a timestamp checked, pass the timestamps as input, not the whole frame

murillo128 (changes requested):

include/TimeStampChecker.h
```diff
@@ -0,0 +1,95 @@
...
+	{
+		// Check if it is just started
+		if (lastRecvTime == 0 && lastTimeStamp == 0)
+		{
+			lastTimeStamp = timestamp;
+			lastRecvTime = recvTimeMs;
+			return CheckResult::Valid;
+		}
+		
+		auto durationOnTs = (timestamp - lastTimeStamp) * 1000 / clock;
```
murillo128, line 51: either auto is an int64, and then we have propblems if timestamp - lastTimestamp value is too big, or it is and uint64_t and we have problems when timestamp < lastTimestamp.

harryz2000: Yes. Missed this one. Will change to get the diff for whichever is larger.

harryz2000: Changed to use a function to always get positive diff.

harryz2000:
The issue wasn't reproduced in new dev server versions. Close for now.

murillo128 (approved):
lgtm

PR #236, opened 2023-12-13
log error if we ever accidentally reuse 5-tuples

mildsunrise:
if the user accidentally violates the assumption that 5-tuples be unique among transports, log an error.
this mitigates the confusing "DTLS write errors" (and such) that come afterwards due to packets being routed to old transports.


murillo128:
@mildsunrise ping?

mildsunrise:
> I don't like to assign a meaning to an abstract interface just because we happen to use it in a specific way today.

I figured you'd say something like this :sweat_smile:

> If we need to identify the transport somehow, let's add it implicity as an id or similar to the ICECandidate, if it is only for logging we can use the `inserted` value and print a message

very reasonable, I've done just that. I used the transport's username as ID, since it's what we're using already.

> On the future, we should make this method to return a promise and propagate the error to js

yeah, i just preferred not to spend time on this now. I can do that if you want.

murillo128 (approved):
lgtm

PR #237, opened 2023-12-22
Remove RTP header size from media bitrate calculation

harryz2000:
The RTP header size was taken account into the audio/video bitrate calculation, which results in the stats higher than actual media bitrate.

### DIFF
--- a/include/rtp/RTPIncomingSource.h
+++ b/include/rtp/RTPIncomingSource.h
@@ -75,12 +75,12 @@ struct RTPIncomingSource : public RTPSource
 	DWORD ExtendTimestamp(DWORD timestamp);
 	DWORD RecoverTimestamp(DWORD timestamp);
 	
-	void Update(QWORD now,DWORD seqNum,DWORD size,const std::vector<LayerInfo> &layerInfos, bool aggreagtedLayers, const std::optional<struct VideoLayersAllocation>& videoLayersAllocation);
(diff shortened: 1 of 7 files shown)


murillo128:
I don't think we should change the current `totalBytes` and `bitrate` stats, as they are used for multiple things, including bwe and layer selection.

I think we should pass both payload size and rtp header size to the `RTPSource` in two different parameters of the `Update()` method and then add another accumulator for either the payload or the rtp header size (don't think we need both), and then add  `payloadBytes` or `rtpHeaderBytes` and `mediaBitrate`  to the stats

harryz2000:
> I don't think we should change the current `totalBytes` and `bitrate` stats, as they are used for multiple things, including bwe and layer selection.
> 
> I think we should pass both payload size and rtp header size to the `RTPSource` in two different parameters of the `Update()` method and then add another accumulator for either the payload or the rtp header size (don't think we need both), and then add `payloadBytes` or `rtpHeaderBytes` and `mediaBitrate` to the stats

I also thought the stats might be useful for the code. But shouldn't bwe (layer selection) use mediaBitrate as well? For SRT/RTMP, it looks the pure media bitrate is used.

Plus, if we add new stats, I think we would need add them to SRT/RTMP as well. Then would need dash API change to use the new stats.

murillo128:
Lets keep `totalBytes` as it is currently (media+rtp verhead), `bitrate` for media bitrate and add a new `totalBitrate` for media + rtp overhead)

harryz2000:
> Lets keep `totalBytes` as it is currently (media+rtp verhead), `bitrate` for media bitrate and add a new `totalBitrate` for media + rtp overhead)

murillo128 (approved):
lgtm

PR #238, opened 2024-01-02
Add resolution stats for vp9 and av1

harryz2000:
Note AV1 resolution stats needs dependency descriptor available in RTP packets.


src/av1/AV1Depacketizer.cpp
```diff
@@ -138,6 +138,8 @@ MediaFrame* AV1Depacketizer::AddPacket(const RTPPacket::shared& packet)
 		AddPayload(packet->GetMediaData(), packet->GetMediaLength());
 	}
 
+	packet->SetWidth(layer.width);
+	packet->SetHeight(layer.height);
```
murillo128, line 142: Do it in the DependencyDescriptorLayerSelector instead

murillo128 (changes requested):

murillo128 (approved):
lgtm

PR #239, opened 2024-01-03
Parse AV1 OBU in case dependency descriptor is not present for resolution stats

harryz2000:


src/DependencyDescriptorLayerSelector.cpp
```diff
@@ -1,6 +1,8 @@
 #include <vector>
 
 #include "DependencyDescriptorLayerSelector.h"
+#include "av1/AV1.h"
```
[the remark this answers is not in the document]
harryz2000: I tried to use conditions in the class for AV1. But yes, using a separate class would be cleaner.

src/av1/AV1LayerSelector.cpp
```diff
@@ -0,0 +1,55 @@
...
+	uint32_t obuSize = reader.GetLeft();
+	if (header->W != 1)
+	{
+		obuSize = reader.DecodeLev128();
+	}
+	
+	// Not enough data
+	if (obuSize > reader.GetLeft()) return infos;
+	
+	auto info = GetObuInfo(reader.PeekData(), obuSize);
```
murillo128, line 40: we would need to refactor this into ObuInfo::Parse to be consistent with the rest of the objects

murillo128: In the refactor, we should pass the reader, so we don't have to store the `payload` pointer on the `ObuInfo`

murillo128 (changes requested):

murillo128:
lgtm, we can merge and do the refactor in an upcoming PR

PR #242, opened 2024-01-18
prevent duplication of candidates with same remote ip:port

murillo128:


mildsunrise:
context for anyone who's reading this:

#236 was incomplete because I was only checking candidates added through the JS interface. this PR fixes that and checks candidates learnt by the ICE layer itself, too. but the `return` also fixes an older memory safety issue that would occur if we called `ActivateRemoteCandidate` on a connection/transport that does not own the candidate, as the candidate could eventually be removed and this transport would still hold a (dangling) reference to it through its `active` candidate pointer.

mildsunrise (approved):
lgtm

PR #243, opened 2024-01-24
DRAFT: DIOS-3710 Fixed duplicate frame returned from VideoPipe

bcostdolby:


bcostdolby:
There are a lot of other similar places I see in the code. The immediate ones are in pipevideoinput.cpp pipevideooutput.cpp that are almost identical. I assume there are probably a lot of other cases as well. I am going to probably create a separate PR to fix these in other locations but this is required to unlock the VOD streaming team.

bcostdolby:
The only other change I considered was to set cancelledGrab=0 in StartVideoCapture() and Init() in case the object is reused will prevent cancellation of the very first frame the next time around

src/VideoPipe.cpp
```diff
@@ -125,31 +125,63 @@ VideoBuffer::const_shared VideoPipe::GrabFrame(uint32_t timeout)
...
 	}
 
 	//Reset new image flag
 	imgNew=0;
 
+	// If this was true, handle img as normal (may exist or may not) and clear this flag so we dont cancel the next one as well
+	cancelledGrab=0;
+
 	//Get current image
 	videoBuffer = imgBuffer[imgPos];
```
murillo128, line 180: if the intention is to clear the image, should we move the shared pointer instead?
```suggestion
	videoBuffer = std::move(imgBuffer[imgPos]);
```

src/VideoPipe.cpp
```diff
@@ -125,31 +125,63 @@ VideoBuffer::const_shared VideoPipe::GrabFrame(uint32_t timeout)
...
 			//wait
-			int ret = pthread_cond_timedwait(&newPicCond,&newPicMutex,&ts);
-			if (ret && ret!=ETIMEDOUT)
-				Error("-VideoPipe cond timedwait error [%d,%d]\n",ret,errno);
-		} else {
+			ret = pthread_cond_timedwait(&newPicCond,&newPicMutex,&ts);
+			if (ret == ETIMEDOUT)
+			{
+				pthread_mutex_unlock(&newPicMutex);
+				return videoBuffer;
```
murillo128, line 158: I assume that this is the behavior change, return an empty buffer if timeout, right? if so, make it splicit with a comment

bcostdolby:
We cancelled this work as it isnt required for now and will require other changes to the encode side of the transcoder to integrate without changing fps.

PR #247, opened 2024-02-09
Add thread id in C++ logs to help debugging

harryz2000:


include/log.h
```diff
@@ -84,9 +88,10 @@ inline int Log(const char *msg, ...)
 		va_list ap;
 		gettimeofday(&tv,NULL);
 #if defined(__linux__)
+		pid_t tid = gettid();
 		char name[16];
 		pthread_getname_np(pthread_self(), name,sizeof(name));
-		printf("[%-16s][%.10ld.%.3ld][LOG]", name, (long)tv.tv_sec, (long)tv.tv_usec / 1000);
+		printf("[%-16s][%-8d][%.10ld.%.3ld][LOG]", name, tid, (long)tv.tv_sec, (long)tv.tv_usec / 1000);
```
murillo128, line 94: can we use hexadecimal? should be `%-4x` right?

harryz2000: We could use 0x%-4x. Before that, my preference would be to combine the thread name and ID so it could be another way to make it shorter (see below). The thread ID is mostly printed out in dec format. One benefit the PID (could be same as thread ID) which could be seen by "top" command in dec format, if there is just one thread. Please have a look. I could change it to hex if you still prefer. 

`[publisher|1097108   ][1707785732.457][LOG] ......`

harryz2000: Changed to 0x%-4x

include/log.h
```diff
@@ -84,9 +88,10 @@ inline int Log(const char *msg, ...)
 		va_list ap;
 		gettimeofday(&tv,NULL);
 #if defined(__linux__)
+		pid_t tid = gettid();
 		char name[16];
 		pthread_getname_np(pthread_self(), name,sizeof(name));
-		printf("[%-16s][%.10ld.%.3ld][LOG]", name, (long)tv.tv_sec, (long)tv.tv_usec / 1000);
+		printf("[%-16s][0x%-4x][%.10ld.%.3ld][LOG]", name, tid, (long)tv.tv_sec, (long)tv.tv_usec / 1000);
 #else
 		printf("[0x%lx][%.10ld.%.3ld][LOG]", (long)pthread_self(), (long)tv.tv_sec, (long)tv.tv_usec / 1000);
```
murillo128, line 96: even if it will provide no extra info, please add it also here so we have logs vertically alligned

harryz2000: Do you mean change this line?
```suggestion
		printf("[0x%lx][0x%-4x][%.10ld.%.3ld][LOG]", (long)pthread_self(), tid, (long)tv.tv_sec, (long)tv.tv_usec / 1000);
```
This wouldn't be used when __linux__ is defined, would it? Plus, the gettid() is only applicable to Linux.

> Function: pid_t gettid (void)(https://sourceware.org/glibc/manual/html_mono/libc.html#index-gettid)
> Preliminary: | MT-Safe | AS-Safe | AC-Safe | See [POSIX Safety Concepts](https://sourceware.org/glibc/manual/html_mono/libc.html#POSIX-Safety-Concepts).
> 
> The gettid function returns the thread ID of the current thread. The returned value is obtained from the Linux kernel and is not subject to caching. See the discussion of thread IDs above, especially regarding reuse of the IDs of threads which have exited.
> 
> This function is specific to Linux.

murillo128: ups, my bad, somehow I thought it was a normal if..

murillo128 (approved):
lgtm

PR #250, opened 2024-02-20
DIOS-3710 Fix duplicate timestamps in transcoding

bcostdolby:


bcostdolby:
I plan to document to need for this and how it works on confluence so it may be worth reading that once it is done before a code review. 

I also have a number of tests to write still and have only done some manual testing so far.

include/OverwritingFIFO.h
```diff
@@ -0,0 +1,93 @@
...
+ * File:   OverwritingFIFO.h
+ * Author: Brendon Costa
+ */
+
+#ifndef OVERWRITINGFIFO_H
+#define	OVERWRITINGFIFO_H
+
+// FIFO queue implementation that when PushBack() called and it is full, instead of failing to push will overwrite the oldest data in the queue
+template <typename DataT>
+class OverwritingFIFO
```
harryz2000, line 11: This class might not be needed if CircularQueue could be used. https://github.com/medooze/media-server/blob/master/include/CircularQueue.h

src/VideoEncoderWorker.cpp
```diff
@@ -174,182 +161,228 @@ int VideoEncoderWorker::Encode()
...
+	// the encoded frame timeline, what the source picture timestamp is and the fps of the ingress vs encoded video
+	for (
+		uint64_t nextEncodedTimestamp = ToEncodedTimestamp(encodedFrames);
+		nextEncodedTimestamp <= pic->GetTimestamp();
+		nextEncodedTimestamp = ToEncodedTimestamp(encodedFrames)
+		)
+	{
+		// We will duplicate the encoded frame from the nearest frame in time.
+		uint64_t midPoint = (lastSourceFrame->GetTimestamp() + pic->GetTimestamp()) / 2;
+		EncodeFrame(nextEncodedTimestamp < midPoint ? lastSourceFrame : pic, nextEncodedTimestamp);
```
harryz2000, line 287: Couldn't this cause a frame encoded multiple times? Or isn't it an issue?

harryz2000:
Generally I think the change to use a queue-like buffer and calculating timestamps purely basing on incoming timestamps would make it more stable. There are some details that potentially have issues though. I would do further review once it was considered ready to merge.

src/VideoEncoderWorker.cpp
bcostdolby: Yes. That is what happens if changing fps. I.e. When fps source is say 30fps and destination is 60fps, then this can indeed encode the source twice to produce the two destination frames as expected.

Likewise, if the source is 60fps and destination 30fps then we will encode half the frames and skip some.

Additionally it should correctly select how to change the rate for any source/dest combination.

The key difference from the old code is that in the new code the frames encoded are selected deterministically based on timestamps and not based on when the code executed what is sampled on the buffer at the time it executed.

include/OverwritingFIFO.h
bcostdolby: fixed

src/VideoPipe.cpp
```diff
@@ -248,22 +281,25 @@ int VideoPipe::NextFrame(const VideoBuffer::const_shared& videoBuffer)
...
+	
 	//Signal any pending grap operation
 	pthread_cond_signal(&newPicCond);
 
 	//Unlock
 	pthread_mutex_unlock(&newPicMutex);
 
-	return 1;
+	// We will return the size of the queue so it can be used in stats reporting from the decode worker
+	return (int)qsize;
```
murillo128, line 302: should we change the return type to size_t?

src/VideoPipe.cpp
```diff
@@ -248,22 +281,25 @@ int VideoPipe::NextFrame(const VideoBuffer::const_shared& videoBuffer)
...
-	//There is an image available for grabbing
-	imgNew = true;
-
+	// Push the new decoded picture onto the queue (log if dropping it because encoder not keeping up)
+	if (queue.full())
+	{
+		// This might be better as a form of statistic available in the dashboard so we get a better idea 
+		// of transcoder health without flooding the logs when it is overloaded but is udeful for debugging 
+		// right now. We should not see it if everything is operating as expected
+		Debug("-VideoPipe::NextFrame() Video frame queue full with size: %u, pushing: %llu, dropping the oldest frame in the queue: %llu. Is the encoder keeping up?\n", queue.length(), videoBuffer->GetTimestamp(), queue.front()->GetTimestamp());
```
murillo128, line 290: Maybe it is a good idea to add a counter here so we can track the dropped frames

murillo128 (approved):
lgtm, just some minor feedback

src/VideoPipe.cpp
```diff
@@ -231,8 +262,10 @@ void  VideoPipe::CancelGrabFrame()
 	//Lock
 	pthread_mutex_lock(&newPicMutex);
 
-	//No image
-	imgNew = false;
+	// We need to set this to identify we want to exit the cond var loop
+	// on any existing or next call to GrabFrame()
+	cancelledGrab = true;
+	queue.clear();
```
murillo128, line 268: Maybe we should not clear the queue on cancel, but check the cancellledGrab flag on the GrabFrame to not return a frame?

src/VideoPipe.cpp
```diff
@@ -125,30 +125,61 @@ VideoBuffer::const_shared VideoPipe::GrabFrame(uint32_t timeout)
...
+		if (ret)
+		{
+			Error("-VideoPipe cond wait error [%d,%d]\n",ret,errno);
 		}
 	}
 
-	//Reset new image flag
-	imgNew=0;
+	// If this was true, handle img as normal (may exist or may not) and clear this flag so we dont cancel the next one as well
+	cancelledGrab=0;
```
murillo128, line 175: In CancelGrab we reset the queue, if we want to make this explicit, i.e. a canceled grab returns no frame (which seems coherent, we should change this to

```suggestion
	if (cancelledGrab)
	{
		//Reset flag	
		cancelledGrab = 0;
		//A canceled grab returns no frame
		return videoBuffer;
	}
```

src/VideoPipe.cpp
bcostdolby: Yep considered that as well before, but went for minimal change and old code was using imgNew=false so I did the equivalent here. I will change now as I think returning null from cancel makes more sense as well

src/VideoPipe.cpp
bcostdolby: I have now added an accumulator to report stats at Debug once a second (if there are drops in that second), and made the frame level report UltraDebug to try and reduce the verbosity.

Should I create a JIRA ticket to add some transcoder related stats to getstatsdb like dropped frames and latency through the transcoder, and look at using perfetto for the performance related stats for profiling decode/encode costs

src/VideoPipe.cpp
bcostdolby: done

src/VideoPipe.cpp
```diff
@@ -242,28 +288,50 @@ void  VideoPipe::CancelGrabFrame()
...
+		{
+			Debug("-VideoPipe::NextFrame() Video frame queue overflowed dropping %u of %u frames per second. Is the encoder keeping up?\n", (unsigned int)droppedFramesAcu.GetAcumulated(), (unsigned int)droppedFramesAcu.GetCount());
+		}
+
+		droppedFramesAcu.Reset(now);
+		lastDroppedReport = now;
+	}
+
+	// We will return the size of the queue so it can be used in stats reporting from the decode worker
+	return qsize;
```
harryz2000, line 337: Looks returning a stats isn't very valuable. Would that be used by some algorithm? Could we return something like whether a frame was discarded, etc.

Also it appears not used at this moment.

src/VideoPipe.cpp
```diff
@@ -125,30 +134,66 @@ VideoBuffer::const_shared VideoPipe::GrabFrame(uint32_t timeout)
...
+		
+		if (ret)
+		{
+			Error("-VideoPipe cond wait error [%d,%d]\n",ret,errno);
 		}
 	}
 
-	//Reset new image flag
-	imgNew=0;
+	if (cancelledGrab)
```
harryz2000, line 183: Just wanted to raise a question. What if cancelledGrab was set after this line. It seems it would cancel the next grab? Should clear the cancelledGrab when enter the waiting?

src/VideoPipe.cpp
bcostdolby: Not used yet, but I have a second commit for later that I would like to report better stats in the decoder worker that use this.

src/VideoPipe.cpp
bcostdolby: This is under the lock and so is the setting of it. So if the cancel grab set to =1 after we release the lock we will cancel the next grab request.  If it happens before we will cancel this request.

mqian2022: I might miss something here. What is the purpose of the cancel grab? Why do we need to cancel the next grab?

harryz2000: Yes. The issue is why it cancels one grab. Looking from the caller code, I could see it seems only to be cancelled when it is stopped, which means it wouldn't need to grab anymore at all. However, the code is clearing the cancelledGrab if it is true, which means the next call to this function may still block. So it seems cancelledGrab shouldn't be cleared. Otherwise, the function may still block while we want to stop the thread. cancelledGrab probably need to be cleared when it is started.

harryz2000: On the other hand, if we just want to cancel current blocking, the `cancelledGrab = false` could be moved to beginning of the lock before the waiting. However, this seems not the use case.

src/VideoPipe.cpp
harryz2000: imho, it would be better to have stats from this class in a unified place rather than passing to the caller. There are already some stats in other places of this class. I think even a function such as getStats() could be added to gather all the stats if needed. Passing it as return value is just a bit strange to me.
Just raise this issue, feel free to ignore if it doesn't matter to people.

PR #254, opened 2024-02-22
CMakeFiles fixes

mildsunrise:
adds missing source files and libraries to cmake, and build + run the legacy tests (I know we want to switch to gtest ASAP, but it would be nice to be able to run the existing tests in the meantime)


murillo128 (approved):
lgtm

PR #253, opened 2024-02-22
h264: full SPS parsing

mildsunrise:
Please review #252 first, as it's a dependency for this PR.

Extends `H264SeqParameterSet` to parse the full descriptor (right now it's not parsing the VUI).

I've basically adapted the code from nal-extractor:
https://github.com/medooze/nal-extractor/blob/main/lib/SPS.ts
https://github.com/medooze/nal-extractor/blob/main/test/SPS.test.ts

output for:

```
64001eacb2010020f3cf80b5060606a50000030001000003003c9f162e48
```

is:

```
[H264SeqParameterSet
	profile_idc=64
	constraint_set0_flag=0
	constraint_set1_flag=0
	constraint_set2_flag=0
	reserved_zero_5bits=0
	level_idc=1e
	seq_parameter_set_id=0
	log2_max_frame_num_minus4=0
	pic_order_cnt_type=2
	log2_max_pic_order_cnt_lsb_minus4=0
	delta_pic_order_always_zero_flag=0
	offset_for_non_ref_pic=0
	offset_for_top_to_bottom_field=0
	num_ref_frames_in_pic_order_cnt_cycle=0
	num_ref_frames=3
	gaps_in_frame_num_value_allowed_flag=0
	pic_width_in_mbs_minus1=31
	pic_height_in_map_units_minus1=31
	frame_mbs_only_flag=1
	mb_adaptive_frame_field_flag=0
	direct_8x8_inference_flag=1
	frame_crop=[
		left_offset=0
		right_offset=6
		top_offset=0
		bottom_offset=6
	]
	vui_parameters=[
		aspect_ratio_info=[
			aspect_ratio_idc=1
		]
		overscan_info=[
			<Not present>
		]
		video_signal_type=[
			video_format=5
			video_full_range_flag=0
			colour_description=[
				colour_primaries=6
				transfer_characteristics=6
				matrix_coefficients=6
			]
		]
		chroma_loc_info=[
			type_top_field=1
			type_bottom_field=1
		]
		timing_info=[
			num_units_in_tick=1
			time_scale=61
			fixed_frame_rate_flag=1
		]
		nal_hrd_parameters=[
			<Not present>
		]
		vcl_hrd_parameters=[
			<Not present>
		]
		pic_struct_present_flag=1
		bitstream_restriction=[
			motion_vectors_over_pic_boundaries_flag=1
			max_bytes_per_pic_denom=0
			max_bits_per_mb_denom=0
			log2_max_mv_length_horizontal=10
[...]

PR #257, opened 2024-02-22
bitstream: fixes

mildsunrise:
Fixes several issues in bitstream.h (BitReader and BitWriter):

 - in C++, [it is undefined behavior to do `dword >> 32` or `dword << 32`](https://stackoverflow.com/a/18918340/710951). And we are doing that in several points of the code. This is one of the UBs that hardly leads to crashes, but it did lead to an actual bug, which is observed in #253 (see description). In `Get()`, when we combine cache with ret:

   https://github.com/medooze/media-server/blob/6b369a170791bf25f80b056e6f83855e5a5e6291/include/bitstream.h#L91

   This assumes that the lower bits of `cache` are zero. Which is guaranteed since `SkipCached()` does a right shift:

   https://github.com/medooze/media-server/blob/6b369a170791bf25f80b056e6f83855e5a5e6291/include/bitstream.h#L268

   ...except when `n` is 32, in which case this invokes UB. In most implementations / architectures this will result in leaving `cache` unchanged rather than zeroed.

 - `GetVariableLengthUnsigned` makes no sense. It is an AV1 primitive (essentially identical to H.264 `ue(v)`) and it should be implemented as: 
Screenshot from 2024-02-22 21-29-42

 - `Get()` doesn't set error condition if there are not enough bits to read (or rather, it only does that in specific cases).

### DIFF
--- a/include/bitstream.h
+++ b/include/bitstream.h
@@ -91,6 +91,9 @@ public:
 			ret = cache >> (32-n);
 			//Cache next
 			Cache();
+			//Check available
+			if (cached<a)
+				error = true;
 			//Get the remaining
 			ret =  ret | GetCached(a);
 		} else if (n) {
@@ -171,19 +174,21 @@ public:
 		return (v << 1) - m + extra_bit;
 	}
 
-	inline uint64_t GetVariableLengthUnsigned()
+	inline uint32_t GetVariableLengthUnsigned()
 	{
 		int leadingZeros = 0;
-		while (Get(1) && !error)
+		while (1)
 		{
 			bool done = Get(1);
+			if (error)
+				return UINT32_MAX;
 			if (done)
 				break;
 			leadingZeros++;
 		}
 		if (leadingZeros >= 32)
-			return (1lu << 32) - 1;
-		uint64_t value = Get(leadingZeros);
+			return UINT32_MAX;
+		uint32_t value = Get(leadingZeros);
 		return value + (1lu << leadingZeros) - 1;
 	}
 public:
@@ -265,7 +270,10 @@ public:
 		//Check length
 		if (!n) return;
 		//Move
-		cache = cache << n;
+		if (n < 32)
+			cache = cache << n;
+		else
+			cache = 0;
 		//Update cached bytes
 		cached -= n;
 
@@ -414,6 +422,8 @@ public:
 			//Increase cached
 			cached = b;
 		} else {
+			//Don't invoke UB if empty
+			if (!n) return v;
 			//Add to cache
 			cache = (cache << n) | (v & (0xFFFFFFFF>>(32-n)));
 			//Increase cached
--- a/include/bitstream.h
+++ b/include/bitstream.h
@@ -83,17 +83,16 @@ public:
 	inline DWORD Get(DWORD n)
 	{
 		DWORD ret = 0;
-		if (n>cached)
-		{
+		if (n>32) {
+			//We can't use exceptions so set error flag
+			error = true;
+		} else if (n>cached){
 			//What we have to read next
 			BYTE a = n-cached;
 			//Get remaining in the cache
 			ret = cache >> (32-n);
 			//Cache next
 			Cache();
-			//Check available
-			if (cached<a)
-				error = true;
 			//Get the remaining
 			ret =  ret | GetCached(a);
 		} else if (n) {
@@ -112,8 +111,10 @@ public:
 
 	inline void Skip(DWORD n)
 	{
-		if (n>cached)
-		{
+		if (n>32) {
+			//We can't use exceptions so set error flag
+			error = true;
+		} else if (n>cached) {
 			//Get what is left to skip
 			BYTE a = n-cached;
 			//Cache next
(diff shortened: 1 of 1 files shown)


mildsunrise:
More issues that I have:

- `GolombExpDecoder::Decode()` invokes UB with certain user input.
- `BitWritter::Put()` does not signal error if there isn't enough space in buffer (at least, not immediately).
- It is possible to underflow `cached`, for example by reading more bits than present. this can then cause `Left()` to underflow as well.

include/bitstream.h
```diff
@@ -91,6 +91,9 @@ class BitReader
 			ret = cache >> (32-n);
 			//Cache next
 			Cache();
+			//Check available
+			if (cached<a)
```
murillo128, line 95: shouldn't this check be done inside `GetChached()` ?

include/bitstream.h
```diff
@@ -265,7 +270,10 @@ class BitReader
 		//Check length
 		if (!n) return;
 		//Move
-		cache = cache << n;
+		if (n < 32)
+			cache = cache << n;
+		else
+			cache = 0;
 		//Update cached bytes
 		cached -= n;
```
murillo128, line 278: shouldn't we check that `n <= cached`?

include/bitstream.h
```diff
@@ -414,6 +422,8 @@ class BitWritter{
 			//Increase cached
 			cached = b;
 		} else {
+			//Don't invoke UB if empty
+			if (!n) return v;
```
murillo128, line 426: should we put this check on the else?

include/bitstream.h
```diff
@@ -414,6 +422,8 @@ class BitWritter{
 			//Increase cached
 			cached = b;
 		} else {
```
murillo128, line 424:
```suggestion
		} else if (n) {
```

murillo128 (changes requested):

include/bitstream.h
mildsunrise: `GetCached()` is called from two places, and only this call site needs the check. but if you prefer to move it there for clarity, we can do that too

include/bitstream.h
mildsunrise: this is the issue I pointed above (It is possible to underflow cached, for example by reading more bits than present), so yeah let's fix that too

include/bitstream.h
mildsunrise: the idea was to skip the FlushCache() call afterwards, but yes, we can also do that if you prefer

include/bitstream.h
mildsunrise: applied

include/bitstream.h
murillo128: better do it in the inner most layer always

include/bitstream.h
murillo128: we can put it at the begginging of the function, or as a first clause of the if-else-if

murillo128 (approved):
lgtm with just to minor comments, feel free to merge it when you are confortable with it

bcostdolby:
Overall this class makes some basic assumptions on its usage that are not enforced which seems to allow these kinds of issues (especially if needing to decode some kind of variable amount of bits).

It seems there is an "error" state on the class which seems to be its way of notifying issues to the caller. Should we also set that in all these cases (in addition to preventing UB)?

I would think we need to enforce the value passed to this is always <=32 in general for all the public functions.

For example even the Get() does a:
ret = cache >> (32 - n)

if n is given as larger than 32 this is also UB.


I would suggest a common check maybe for all public functions (seems they are all public) for places like:
* Get(n) : Where n > 32
* Check(n) : Where n > 32 (I guess covered by the get, but still)
and others...:
* Skip(n)
* Peek(n)
* SkipCached() 
* GetCached()

seems we are good at checking n == 0 but not n > 32.

bcostdolby:
Also GetNonSymmetric() could probably use some comments describing what it is trying to do. The name doesnt really tell us much.

bcostdolby (changes requested):
I added comments to overall ticket, this class needs more changes I think. Actually reading through the comments, it seems Alba identified some of these but maybe they were not pushed?

mildsunrise:
many of the tasks described in the comments of this PR, as well as outstanding comments in diffs, have been addressed now.

I'll make a list of the ones that haven't been addressed yet (please everyone verify I'm not missing anything):

- GolombExpDecoder::Decode() invokes UB with certain user input
- BitWritter::Put() does not signal error if there isn't enough space in buffer (at least, not immediately)
- check for n>32 in Check() too, requested by @bcostdolby (actually I'd change all these checks into asserts :smiling_imp: but I understand if others don't wanna)
- GetNonSymmetric() could probably use some comments describing what it is trying to do

some questions:

> It seems there is an "error" state on the class which seems to be its way of notifying issues to the caller. Should we also set that in all these cases (in addition to preventing UB)?

regarding this, I think we should do the opposite, and convert this file to use exceptions. not sure how others see it

in any case, I'd merge this now and we can open a new PR for further fixes.
also, @bcostdolby you mentioned you added things to "overall ticket", which ticket is that?

murillo128 (approved):
lgtm, let's do the rest of the fixes in a different PR. Probably also introduce exceptions.

murillo128:
+1 to replace error flags by proper exception handling.

Not sure about asserts, I would be worried about crashing on production
because a software bug.

PR #261, opened 2024-02-24
adapt #259 to #260

mildsunrise:
adapt #259 to #260

### DIFF
--- a/src/rtmp/rtmppacketizer.cpp
+++ b/src/rtmp/rtmppacketizer.cpp
@@ -2,6 +2,7 @@
 #include "rtmp/rtmppacketizer.h"
 #include "av1/AV1.h"
 #include "av1/Obu.h"
+#include "h264/H26xNal.h"
 
 
 VideoCodec::Type GetRtmpFrameVideoCodec(const RTMPVideoFrame& videoFrame)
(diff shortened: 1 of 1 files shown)


src/rtmp/rtmppacketizer.cpp
```diff
@@ -205,11 +206,15 @@ std::unique_ptr<VideoFrame> RTMPH26xPacketizer<DescClass, SPSClass, PPSClass, co
 		//Set timestamp
 		set8(sei, 19, videoFrame->GetSenderTime());
 
+		//Escape nal
+		uint8_t seiEscaped[sizeof(sei)*2];
+		auto seiSize = NalEscapeRbsp(seiEscaped, sizeof(seiEscaped), sei, sizeof(sei)).value();
+
 		//Append nal
-		auto ini = frame->AppendMedia(sei, sizeof(sei));
+		auto ini = frame->AppendMedia(seiEscaped, seiSize);
```
murillo128, line 214: seiSize is an optional, wouldn't this be? 
```suggestion
		auto ini = frame->AppendMedia(seiEscaped, *seiSize);
```

src/rtmp/rtmppacketizer.cpp
```diff
@@ -205,11 +206,15 @@ std::unique_ptr<VideoFrame> RTMPH26xPacketizer<DescClass, SPSClass, PPSClass, co
...
 
 		//Crete rtp packet
-		frame->AddRtpPacket(ini, sizeof(sei), nullptr, 0);
+		frame->AddRtpPacket(ini, seiSize, nullptr, 0);
```
murillo128, line 217: same 
```suggestion
		frame->AddRtpPacket(ini, *seiSize, nullptr, 0);
```

src/rtmp/rtmppacketizer.cpp
murillo128: also, we should check returned value

murillo128 (approved):

src/rtmp/rtmppacketizer.cpp
murillo128: ok, missed the `value()`

PR #266, opened 2024-03-04
Passing SSRC, change threading 

harryz2000:


include/SimulcastMediaFrameListener.h
```diff
@@ -41,8 +41,8 @@ class SimulcastMediaFrameListener :
 private:
 	void ForwardFrame(VideoFrame& frame);
 
-	void Push(std::unique_ptr<VideoFrame>&& frame);
-	void Enqueue(std::unique_ptr<VideoFrame>&& frame);
+	void Push(std::shared_ptr<VideoFrame>&& frame);
```
murillo128, line 44: why do we need a `shared_ptr` ?

harryz2000: Because now the Push function was called in the TimeService::Async, which accept an std::function as argument. Unfortunately, to cast to std::function, the object needs to be copy-able. The std::unique_ptr is not copy-able. Another option could be to templatize the Async function, which would be a quite big change and probably not feasible at this point.

harryz2000: Using a raw pointer when passing to the lambda (then std::function) would also compile. However, it would potentially cause memory leak.

src/MediaFrameListenerBridge.cpp
```diff
@@ -168,11 +168,15 @@ void MediaFrameListenerBridge::RemoveListener(RTPIncomingMediaStream::Listener*
 
 void MediaFrameListenerBridge::onMediaFrame(DWORD ignored, const MediaFrame& frame)
 {
-	timeService.Async([=, frame = std::shared_ptr<MediaFrame>(frame.Clone())] (auto now){
+	std::shared_ptr<MediaFrame> cloned(frame.Clone());
+	cloned->SetSSRC(ssrc);
```
murillo128, line 172: if `MediaFrame::Clone` doesn't clone the SSRC, please fix it instead

src/VideoPipe.cpp
```diff
@@ -255,7 +255,7 @@ VideoBuffer::const_shared VideoPipe::GrabFrame(uint32_t timeout)
 		if (videoBuffer->HasTimestamp())
 			resized->SetTimestamp(videoBuffer->GetTimestamp());
 		if (videoBuffer->HasTime())
-			resized->SetTime(videoBuffer->HasTime());
+			resized->SetTime(videoBuffer->HasTime() ? videoBuffer->GetTime() : getTime());
```
murillo128, line 258: Please add a different PR

src/SimulcastMediaFrameListener.cpp
```diff
@@ -124,10 +124,12 @@ void SimulcastMediaFrameListener::onMediaFrame(DWORD ssrc, const MediaFrame& fra
 		return;
 
 	//Get cloned video frame
-	auto cloned = std::unique_ptr<VideoFrame>((VideoFrame*)frame.Clone());
+	std::shared_ptr<VideoFrame> cloned(static_cast<VideoFrame*>(frame.Clone()));
 	cloned->SetSSRC(ssrc);
 
-	Push(std::move(cloned));
+	timeService.Async([this, cloned = std::move(cloned)](std::chrono::milliseconds) mutable {
```
murillo128, line 130: If you need to make sure that the `Push` method is called in the `timeService` thread, better move the `timeService.Async` method inside the `Push`method

murillo128 (changes requested):

src/SimulcastMediaFrameListener.cpp
harryz2000: OK

src/MediaFrameListenerBridge.cpp
harryz2000: Right. Will check.

src/VideoPipe.cpp
harryz2000: Actually this has to be fixed so that the SimulcastMediaFrameListener could work. Do you still prefer to have a separate PR?

src/MediaFrameListenerBridge.cpp
harryz2000: The SSRC here is not using the frame's own ssrc. I fixed the clone() function. However, still need to manually set here.

harryz2000:
@murillo128 This PR is dependency of other PRs. Are you OK for this one?

src/VideoPipe.cpp
murillo128: I think it is better to not have "jumbo" commits, so it is not about the PR itself but having lots of changes into a single commit,

src/MediaFrameListenerBridge.cpp
```diff
@@ -168,7 +168,9 @@ void MediaFrameListenerBridge::RemoveListener(RTPIncomingMediaStream::Listener*
...
+	timeService.Async([=, frame = std::move(cloned)] (auto now){
```
murillo128, line 173: I see,  maybe it is better to refector a bit to avoid the impresion that it is a problem in the `Clone()` function but that we need to override the ssrc of the incoming ssrc to the oitgoing ssrc
```suggestion
	timeService.Async([=, frame = std::shared_ptr<MediaFrame>(frame.Clone())] (auto now){
		//Update outgoing ssrc before multiplexing
		frame->SetSSRC(ssrc);
```

murillo128 (approved):
lgtm

PR #271, opened 2024-03-20
VideoDecoder: accept MediaFrames, not RTPPackets

mildsunrise:
this commit makes VideoDecoderWorker implement `MediaFrame::Listener` interface instead of `RTPIncomingMediaStream::Listener`. unless I'm missing something, the change looks pretty straightforward to me since the interfaces are very similar and map well.

this change drops the loss detection logic and the partial frame retaining logic. loss detection is only used to update the `waitIntra` flag, which is just used for logging.

these two features (loss detection and partial frame retaining) belong in the depacketizers, not here. if they are added in the future to the depacketizers, we can add boolean attributes to MediaFrame to signal frame loss or partial frames, and then all that would be needed in VideoDecoderWorker is to check those flags and update `waitIntra` as needed.

### DIFF
--- a/src/VideoDecoderWorker.cpp
+++ b/src/VideoDecoderWorker.cpp
@@ -84,7 +84,6 @@ int VideoDecoderWorker::Decode()
 	QWORD		frameTime = (QWORD)-1;
 	DWORD		frameClockRate = (DWORD)-1;
 	QWORD		frameSenderTime = 0;
-	DWORD		lastSeq = RTPPacket::MaxExtSeqNum;
 	bool		waitIntra = false;
 
 	Log(">VideoDecoderWorker::Decode()\n");
@@ -106,11 +105,10 @@ int VideoDecoderWorker::Decode()
 			continue;
 		
 		//Get extended sequence number and timestamp
-		DWORD seq = packet->GetExtSeqNum();
-		QWORD ts = packet->GetExtTimestamp();
+		QWORD ts = packet->GetTimestamp();
 		QWORD time = packet->GetTime();
 		DWORD clockRate = packet->GetClockRate();
-		QWORD senderTime = packet->HasAbsoluteCaptureTime() ? packet->GetAbsoluteCaptureTime() : 0;
+		QWORD senderTime = packet->GetSenderTime();
 
 		//If we don't have codec
 		if (!videoDecoder || (packet->GetCodec()!=videoDecoder->type))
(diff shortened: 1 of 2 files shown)


src/VideoDecoderWorker.cpp
```diff
@@ -265,21 +200,18 @@ int VideoDecoderWorker::Decode()
 	return 0;
 }
 
-void VideoDecoderWorker::onRTP(const RTPIncomingMediaStream* stream, const RTPPacket::shared& packet)
+void VideoDecoderWorker::onMediaFrame(const MediaFrame& frame_)
```
murillo128, line 203:
```suggestion
void VideoDecoderWorker::onMediaFrame(const MediaFrame& frame)
```

src/VideoDecoderWorker.cpp
```diff
@@ -265,21 +200,18 @@ int VideoDecoderWorker::Decode()
...
-	packets.Add(packet->Clone());
-}
-
-void VideoDecoderWorker::onEnded(const RTPIncomingMediaStream* stream)
-{
-	//Cancel packets wait
-	packets.Cancel();
-}
+	//Clone it
+	auto frame = std::shared_ptr<MediaFrame>(frame_.Clone());
```
murillo128, line 206: do the check about the frame type first on 
```suggestion
	//Check it's a video frame
	if (frame->GetType() != MediaFrame::Type::Video)
	{
		Warning("-VideoDecoderWorker::onMediaFrame(): got %s frame [this: %p]\n", MediaFrame::TypeToString(frame->GetType()), this);
		return;
	}
	//Clone it
	auto cloned = std::shared_ptr<MediaFrame>(frame.Clone());
```

murillo128 (approved):
lgtm, but we should clean the VideoDecoder interface and implementations as well

src/VideoDecoderWorker.cpp
```diff
@@ -92,89 +88,40 @@ int VideoDecoderWorker::Decode()
...
-		}
-		
-		//Update frame timestamp
-		frameTimestamp = ts;
-		frameTime = time;
-		frameClockRate = clockRate;
-		
 		//Decode packet
-		if(!videoDecoder->DecodePacket(packet->GetMediaData(),packet->GetMediaLength(),lost,packet->GetMark()))
+		if(!videoDecoder->Decode(encFrame->GetData(), encFrame->GetLength()))
```
murillo128, line 121: we should check the `VideoDecoder` interface to receive a `const VideoFrame::shared&` instead

mildsunrise:
> lgtm, but we should clean the VideoDecoder interface and implementations as well

yes, I was planning to do that afterwards (and also replicate the changes in audio, too). this was just a proof of concept. let me test it and I'll do the VideoDecoder refactor

murillo128:
let's prioritize this PR as I need to make other changes to the VideoEncoder which will benefit for this refactor

mildsunrise:
@bcostdolby you said you would take over on this, did you start work yet? if not, I can continue working on the branch once I'm done with the initial version of the filter

@murillo128 which changes are you planning to make on VideoEncoder?

bcostdolby:
> @bcostdolby you said you would take over on this, did you start work yet? if not, I can continue working on the branch once I'm done with the initial version of the filter
> 
> @murillo128 which changes are you planning to make on VideoEncoder?

No. I have not started on this issue yet. If you have time now then please take it.

murillo128:
I have continued the work here: https://github.com/medooze/media-server/pull/281

PR #272, opened 2024-03-27
DIOS-4895 Add detailed log messages to support debugging common RTMP/SRT connection issues

bcostdolby:


src/rtmp/rtmpconnection.cpp
```diff
@@ -898,6 +898,18 @@ void RTMPConnection::ProcessCommandMessage(DWORD streamId,RTMPCommandMessage* cm
 
 	//Log
 	Log("-RTMPConnection::ProcessCommandMessage() [streamId:%d,name:\"%ls\",transId:%ld]\n",streamId,name.c_str(),transId);
+	if (params && params->GetType() != AMFData::ValueType::Null)
```
harryz2000, line 901: Personally I think all the logic that used for debugging should be only enabled when debug level log is enabled. Otherwise, it costs performance in release code. Maybe consider use the following pattern?

```
if (Logger.IsDebugEnabled())
{
...
}
```

bcostdolby: done

harryz2000 (approved):
lgtm

PR #275, opened 2024-04-08
DIOS-5077 Give warning level logs their own config not tied to debug

bcostdolby:
…level

### DIFF
--- a/include/log.h
+++ b/include/log.h
@@ -36,6 +36,11 @@ public:
 		return getInstance().log;
 	}
 
+	static bool IsWarningEnabled()
+	{
+		return getInstance().warning;
+	}
+
 	static bool EnableDebug(bool debug)
 	{
 		return getInstance().debug = debug;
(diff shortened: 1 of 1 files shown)


include/log.h
```diff
@@ -177,7 +189,7 @@ inline int Debug(const char *msg, ...)
 
 inline int Warning(const char *msg, ...)
 {
-	if (Logger::IsDebugEnabled())
+	if (Logger::IsWarningEnabled())
```
harryz2000, line 192: It looks we should enable higher level logs when lower level is enabled. Shouldn't we?

harryz2000: OK. Maybe I was wrong. The levels look independent except the UltraDebug and Debug.

harryz2000 (approved):
lgtm

bcostdolby:
This is a very small change no changes were required so merging now so I can properly update the other repos to use it (need the submodule hash updated).

PR #280, opened 2024-05-13
Only update stats when resolution is valid

harryz2000:
This avoid the stats being overwritten by frames whose resolution is not set

### DIFF
--- a/src/MediaFrameListenerBridge.cpp
+++ b/src/MediaFrameListenerBridge.cpp
@@ -228,6 +228,12 @@ void MediaFrameListenerBridge::onMediaFrame(DWORD ignored, const MediaFrame& fra
 				width = videoFrame->GetWidth();
 				height = videoFrame->GetHeight();
 			}
+			else if (width != 0 && height != 0)
+			{
+				// Set as per previous width/height
+				videoFrame->SetWidth(width);
+				videoFrame->SetHeight(height);
+			}
 
 			// Increase bframes
 			accumulatorBFrames.Update(now.count(), videoFrame->IsBFrame() ? 1 : 0);
@@ -400,6 +406,17 @@ void MediaFrameListenerBridge::onMediaFrame(DWORD ignored, const MediaFrame& fra
 
 				//TODO: move out of here
 				VideoLayerSelector::GetLayerIds(packet);
+				
+				// Set frame resolution if needed
+				if ((video->GetWidth() == 0 || video->GetHeight() == 0) && 
+				    (packet->GetWidth() != 0 && packet->GetHeight() != 0))
+				{
+					video->SetWidth(packet->GetWidth());
+					video->SetHeight(packet->GetHeight());
+					
+					width = video->GetWidth();
+					height = video->GetHeight();
+				}
 
 				//Get media frame target bitrate or hint
 				auto targetBitrate = video->GetTargetBitrate() ? video->GetTargetBitrate() : targetBitrateHint;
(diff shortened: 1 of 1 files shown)


src/MediaFrameListenerBridge.cpp
```diff
@@ -221,9 +221,13 @@ void MediaFrameListenerBridge::onMediaFrame(DWORD ignored, const MediaFrame& fra
...
 			//Convert
 			auto  videoFrame = std::static_pointer_cast<VideoFrame>(frame);
-			//Set width and height
-			width = videoFrame->GetWidth();
-			height = videoFrame->GetHeight();
+			
+			if (videoFrame->GetWidth() != 0 && videoFrame->GetHeight() != 0)
+			{
+				//Set width and height
+				width = videoFrame->GetWidth();
```
bcostdolby, line 228: Why is there no width/height in the first place? 

Concept of change lgtm but surprised it is necessary or even visible. Most video frames have resolution on them so even if one or two of them didnt would probably not see that. I would expect other things to be broken as well unless there are special no-video video frames that exist I dont know about.

harryz2000: The resolution was parsed from current frame data. Not all frame data contain width/height. i.e. it might be only the key frame that contains width/heigh information. We could set frame resolution to be same as one of the previous frames, which would be similar as current way to set the stats.

harryz2000: Actually I can set the width here if it is not valid, which should be no harm. Will add some extra code.

src/MediaFrameListenerBridge.cpp
```diff
@@ -221,9 +221,19 @@ void MediaFrameListenerBridge::onMediaFrame(DWORD ignored, const MediaFrame& fra
...
+				height = videoFrame->GetHeight();
+			}
+			else if (width != 0 && height != 0)
+			{
+				// Set as per previous width/height
+				videoFrame->SetWidth(width);
```
murillo128, line 234: I am not sure we should be modifying the VideoFrame object at all, note that it is not "owned" by the MediaFrameListenerBridge

murillo128 (changes requested):

src/MediaFrameListenerBridge.cpp
harryz2000: Right. I can remove this frame setting. It wouldn't impact current fix and it seems fine it is not set on the frame.

src/MediaFrameListenerBridge.cpp
```diff
@@ -396,6 +400,14 @@ void MediaFrameListenerBridge::onMediaFrame(DWORD ignored, const MediaFrame& fra
 
 				//TODO: move out of here
 				VideoLayerSelector::GetLayerIds(packet);
+				
+				// Set frame resolution if needed
+				if ((video->GetWidth() == 0 || video->GetHeight() == 0) && 
+				    (packet->GetWidth() != 0 && packet->GetHeight() != 0))
+				{
+					width = video->GetWidth();
```
murillo128, line 408: As we have removed setting the width/height on the video frames, we need to use the packet info instead
```suggestion
					width = packet->GetWidth();
```

murillo128 (changes requested):

src/MediaFrameListenerBridge.cpp
```diff
@@ -396,6 +400,14 @@ void MediaFrameListenerBridge::onMediaFrame(DWORD ignored, const MediaFrame& fra
...
+					height = video->GetHeight();
```
murillo128, line 409:
```suggestion
					height = packet->GetHeight();
```

harryz2000: Indeed.

harryz2000:
@murillo128 Please have another look. I think it is good for merge now. thx

murillo128 (approved):
lgtm

PR #282, opened 2024-05-13
Return VideoBuffer to VideoBufferPool when shared pointer is destroyed

murillo128:


murillo128:
Current code was just a factory, it didn't return the object to the pool when it was released

include/VideoBuffer.h
```diff
@@ -122,6 +122,18 @@ class VideoBuffer
 	DWORD   GetClockRate() const	{ return clockRate.value();		}
 	void    SetClockRate(DWORD clockRate) { this->clockRate = clockRate;	}
 
+	void	Reset()
+	{
```
harryz2000, line 126: Is there a reason why planeY, planeU, planeV, width, height were not reset? Probably not necessary, but seems it would be good to reset all of them.

include/VideoBufferPool.h
```diff
@@ -67,7 +67,14 @@ class VideoBufferPool
 
 
 		//We need to create a new one
-		return VideoBuffer::shared(buffer);
+		return VideoBuffer::shared(buffer, [&](auto p) {
+			//Reset it
+			p->Reset();
+			//Ensure we are not overallocating
+			if (maxallocate && pool.size_approx() < maxallocate)
```
harryz2000, line 74: What if the condition was not met? Should we `delete p`?

include/VideoBuffer.h
murillo128: Those values can't be changed, so no need to reset them

include/VideoBufferPool.h
murillo128: good catch!

include/VideoBufferPool.h
```diff
@@ -67,21 +68,31 @@ class VideoBufferPool
...
+			if (maxallocate && pool && pool->size_approx() < maxallocate)
+			{
+				//Reset it
+				p->Reset();
 				//Enqueue it back
-				pool.enqueue(p);
+				pool->enqueue(p);
+			} else {
+				//Delete it
+				delete (buffer);
```
harryz2000, line 82: I think this should be changed to `delete p`. For example, if the shared_ptr.reset(new VideoBuffer) is called, the deleter should not always try to delete the pointer when the shared_ptr was constructed.

harryz2000: other than this, lgtm

include/VideoBufferPool.h
```diff
@@ -67,14 +67,31 @@ class VideoBufferPool
 
 
 		//We need to create a new one
-		return VideoBuffer::shared(buffer);
+		return VideoBuffer::shared(buffer, [=, weak = std::weak_ptr<moodycamel::ConcurrentQueue<VideoBuffer*>>(pool) ](auto p) {
+			//Try to get a reference to the pool, as it may have been already deleted
+			auto pool = weak.lock();
+			//Ensure we are not overallocating
+			if (maxallocate && pool && pool->size_approx() < maxallocate)
```
bcostdolby, line 74: If pool can go out of scope then can this->maxallocated also be out of scope?

include/VideoBufferPool.h
```diff
@@ -67,14 +67,31 @@ class VideoBufferPool
 
 
 		//We need to create a new one
-		return VideoBuffer::shared(buffer);
+		return VideoBuffer::shared(buffer, [=, weak = std::weak_ptr<moodycamel::ConcurrentQueue<VideoBuffer*>>(pool) ](auto p) {
```
murillo128, line 70:
```suggestion
		return VideoBuffer::shared(buffer, [maxallocate = maxallocate, weak = std::weak_ptr<moodycamel::ConcurrentQueue<VideoBuffer*>>(pool) ](auto p) {
```

include/VideoBufferPool.h
```diff
@@ -67,14 +67,31 @@ class VideoBufferPool
...
+			//Ensure we are not overallocating
+			if (maxallocate && pool && pool->size_approx() < maxallocate)
+			{
+				//Reset it
+				p->Reset();
+				//Enqueue it back
+				pool->enqueue(p);
+			} else {
+				//Delete it
+				delete (buffer);
```
murillo128, line 82:
```suggestion
				delete (p);
```

include/VideoBufferPool.h
murillo128: right, I thought about this, but not sure why I concluded that the `=` in the capture was going to make a copy of it. Capturing it explicitly now.

include/VideoBufferPool.h
murillo128: rigth

harryz2000 (approved):
lgtm

PR #283, opened 2024-05-13
DIOS-4552 fix pic rescale rounding issue and unit test

yezhan10:


bcostdolby (changes requested):

src/VideoPipe.cpp
```diff
@@ -221,19 +221,11 @@ VideoBuffer::const_shared VideoPipe::GrabFrame(uint32_t timeout)
...
-		} else {
-			//Check adjusted video size
-			videoWidth = ((uint32_t)(videoBuffer->GetWidth() / scale)) & ~1;
-			videoHeight = ((uint32_t)((videoBuffer->GetHeight() * parDen) / (scale * parNum))) & ~1;
-		}
-
+		//rescaled video height is the height configured through config file
+		videoHeight = scaleResolutionToHeight;
+		//rescale width to retain DAR, plus 1 to round up to nearest even number
+		videoWidth = ((uint32_t)((videoBuffer->GetWidth() * parNum) / (scale * parDen)) + 1) & ~1;
```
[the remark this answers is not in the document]
yezhan10: maybe std::trunc should be used instead of std::round since round(x) returns integral value that is nearest to x, which means there is possibility that x is rounded up in case of fractional part of x is greater than 0.5.  And I guess that converting a float to int as the same effect as trunc operation mathematically.

src/VideoPipe.cpp
```diff
@@ -221,19 +221,11 @@ VideoBuffer::const_shared VideoPipe::GrabFrame(uint32_t timeout)
...
-			videoWidth = ((uint32_t)((videoBuffer->GetWidth() * parNum) / (scale * parDen))) & ~1;
-			videoHeight = ((uint32_t)(videoBuffer->GetHeight() / scale)) & ~1;
-		} else {
-			//Check adjusted video size
-			videoWidth = ((uint32_t)(videoBuffer->GetWidth() / scale)) & ~1;
-			videoHeight = ((uint32_t)((videoBuffer->GetHeight() * parDen) / (scale * parNum))) & ~1;
-		}
-
+		//rescaled video height is the height configured through config file
+		videoHeight = scaleResolutionToHeight;
```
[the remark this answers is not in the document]
yezhan10: sure, i was under impression that scaleResolutionDownBy is not supported anymore.

src/VideoPipe.cpp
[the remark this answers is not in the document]
bcostdolby: Yes you are correct. Keep the code as it is with cast is fine.

I was thinking of it as two separate processes. 

1) Find nearest pixel with (might give something like 122.9999999 instead of 123 so should really result in 123
2) Find nearest even pixel from the pixel accurate width

But that means double rounding which causes issues as you mentioned.

bcostdolby (approved):
lgtm

src/VideoPipe.cpp
```diff
@@ -197,43 +197,41 @@ VideoBuffer::const_shared VideoPipe::GrabFrame(uint32_t timeout)
...
-			videoWidth = ((uint32_t)((videoBuffer->GetWidth() * parNum) / (scale * parDen))) & ~1;
-			videoHeight = ((uint32_t)(videoBuffer->GetHeight() / scale)) & ~1;
-		} else {
-			//Check adjusted video size
-			videoWidth = ((uint32_t)(videoBuffer->GetWidth() / scale)) & ~1;
-			videoHeight = ((uint32_t)((videoBuffer->GetHeight() * parDen) / (scale * parNum))) & ~1;
-		}
-
+		//rescaled video height
+		videoHeight = (uint32_t)(videoBuffer->GetHeight() / scale) & ~1;
```
harryz2000, line 232: Shouldn't his also be rounded to nearest even number?
Maybe worth to create a helper function, i.e.
```
constexpr uint32_t roundToNearstEven(float value)
{
    return uint32_t(value + 1.0f) & ~1;
}
```

src/VideoPipe.cpp
```diff
@@ -197,43 +197,41 @@ VideoBuffer::const_shared VideoPipe::GrabFrame(uint32_t timeout)
...
-	{
-		//Get video height
-		auto videoHeight = videoBuffer->GetHeight();
+	if (scaleResolutionToHeight) {
 
+		auto srcVideoHeight = videoBuffer->GetHeight();
 		//Check if we are allowed to downscale given the input height
-		if ((allowedDownScaling == AllowedDownScaling::SameOrLower && videoHeight<scaleResolutionToHeight)
-			|| (allowedDownScaling == AllowedDownScaling::LowerOnly && videoHeight <= scaleResolutionToHeight))
+		if ((allowedDownScaling == AllowedDownScaling::SameOrLower && srcVideoHeight < scaleResolutionToHeight)
```
murillo128, line 204: I would prefer to keep a single `if` for both `scaleResolutionDownBy` and `scaleResolutionToHeight` would be easy to cause regressions in the future if not.

bcostdolby: Are we keeping support for scaleResolutionDownBy later? I dont think we use it at all.

murillo128: yes we are keeping it

src/VideoPipe.cpp
```diff
@@ -196,44 +196,34 @@ VideoBuffer::const_shared VideoPipe::GrabFrame(uint32_t timeout)
...
+	if (scaleResolutionToHeight || scaleResolutionDownBy) 
 	{
-		//Get pixel aspect ratio
+		float scale = 0.0f;
+		auto srcVideoHeight = videoBuffer->GetHeight();
 		const auto [parNum, parDen] = videoBuffer->GetPixelAspectRatio();
-
-		//Choose which dimension we want to use for adjusting the pixel aspect ratio
-		if (parNum>parDen)
+		if(scaleResolutionToHeight) {
```
bcostdolby, line 205: By the way. Super minor but apparently our style puts opening braces on a new line. 

I have a lint script here locally somewhere I can look for and pass through. It doesnt enforce and lots of code doesnt match it but new code can.

bcostdolby (approved):
Just minor style change otherwise lgtm

PR #286, opened 2024-05-15
Generate medooze-media-server-src as an npm package

Bryan-Lester-Dolby:
This uses a similar strategy as the NodeJS NaN library: https://github.com/nodejs/nan, where our main entrypoint simply returns a relative path to the working directory, allowing us to reference it flexibly as either a direct dependency or a peerDependency without needing to hardcode paths in binding.gyp.

I have left my build hook in, as I anticipate there may be more rapid iteration on the package as we find out special use cases of the libraries that consume it. Right now though, it works with @medooze/gpac-packager: https://github.com/medooze/gpac-packager-node/actions/runs/9103260577/job/25024731525

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -7,7 +7,7 @@ on:
       - 'lester/*'
 
 jobs:
-  publish-to-github-packages:
+  publish-npmjs:
     permissions:
       packages: write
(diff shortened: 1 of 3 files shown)


murillo128 (approved):
lgtm

PR #287, opened 2024-05-22
Add estimateTimeoutTimer so the sse bwe event is fired even if no data is sent

murillo128:


src/DTLSICETransport.cpp
```diff
@@ -3032,6 +3053,14 @@ void DTLSICETransport::EnableSenderSideEstimation(bool enabled)
 	//Update flag
 	this->senderSideEstimationEnabled = enabled;
 
+	//If enabled
+	if (this->senderSideEstimationEnabled)
+		//Start estimation timer
+		estimateTimeoutTimer->Reschedule(EstimationTimeoutInterval, EstimationTimeoutInterval);
```
harryz2000, line 3059: Should we check whether `estimateTimeoutTimer` is valid?

src/DTLSICETransport.cpp
```diff
@@ -3032,6 +3053,14 @@ void DTLSICETransport::EnableSenderSideEstimation(bool enabled)
 	//Update flag
 	this->senderSideEstimationEnabled = enabled;
 
+	//If enabled
+	if (this->senderSideEstimationEnabled)
+		//Start estimation timer
+		estimateTimeoutTimer->Reschedule(EstimationTimeoutInterval, EstimationTimeoutInterval);
+	else
+		//Cancel it
+		estimateTimeoutTimer->Cancel();
```
harryz2000, line 3062: As above

harryz2000 (commented):
Looks good. Just a minor comment.

PR #288, opened 2024-05-31
DIOS-5789 Fix B-frames by adding PTS metadata used when decoders reorder frames

bcostdolby:


include/VideoBuffer.h
```diff
@@ -122,15 +122,24 @@ class VideoBuffer
 	DWORD   GetClockRate() const	{ return clockRate.value();		}
 	void    SetClockRate(DWORD clockRate) { this->clockRate = clockRate;	}
 
+	void SetTSClockRate(uint32_t v) { tsClockRate = v;}
```
murillo128, line 125: do we need a different clockrate for PTS and timestamp?

include/rtp/RTPPacket.h
```diff
@@ -191,6 +191,14 @@ class RTPPacket
 	void SetWidth(uint32_t width)	{ this->width = width;		}
 	void SetHeight(uint32_t height) { this->height = height;	}
 
+
+	void SetTSClockRate(uint32_t v) { tsClockRate = v;}
```
murillo128, line 195: RTP doesn't have PTS

include/VideoBuffer.h
bcostdolby: Not necessarily no. I considered removing this (partly why this is a draft). 

There are times where the clock rate changes, but that is in the packetization and so probably not relevant. I will verify there are no other cases in the code and remove next week if it isn't necessary.

include/video.h
```diff
@@ -119,6 +125,13 @@ class VideoFrame : public MediaFrame
 	void SetBFrame(bool isBFrame) { this->isBFrame = isBFrame; }
 	bool IsBFrame() const { return isBFrame; }
 
+	void SetTSClockRate(uint32_t v) { tsClockRate = v;}
+	void SetPTS(uint64_t v) { pts = v;}
+	void SetDTS(uint64_t v) { dts = v;}
```
murillo128, line 130: Do we need both to add both PTS and DTS  values? What would be the meaning of Timestamp then?

src/VideoPipe.cpp
```diff
@@ -250,6 +250,9 @@ VideoBuffer::const_shared VideoPipe::GrabFrame(uint32_t timeout)
 		if (videoBuffer->HasSenderTime())
 			resized->SetSenderTime(videoBuffer->GetSenderTime());
 
+		resized->SetPTS(videoBuffer->GetPTS());
+		resized->SetTSClockRate(videoBuffer->GetTSClockRate());
+
```
murillo128, line 255: Let's add a CopyTimingInfo to VideoBuffer as well

murillo128 (changes requested):

include/VideoBuffer.h
```diff
@@ -122,15 +122,24 @@ class VideoBuffer
...
+	void SetPTS(uint64_t v) { pts = v;}
```
murillo128, line 126: Not sure if for consistency we should rename `SetPTS()` to `SetPresentationTimestamp()`

include/video.h
```diff
@@ -196,17 +220,25 @@ class VideoDecoder
 
 };
 
-inline void CopyTimingInfo(const VideoFrame::const_shared& videoFrame, VideoBuffer::shared& videoBuffer)
+// Used by decoders after potentially reordering so needs to copy PTS from source
+// to the decoded frame timestamp
+//
+// Unlike SetTimingInfo() this takes into account the change in meaning of "timestamp" from a DTS to a PTS
+inline void CopyTimingInfoFromEncodedToDecoded(const VideoFrame::const_shared& videoFrame, VideoBuffer::shared& videoBuffer)
```
murillo128, line 227: Could we try to get a better naming based on what the function does and not on where it is used?

include/rtp/RTPPacket.h
bcostdolby: correct. this came from when I looked at this before and we had packetization between the ingest and transcoder. fixing this now.

include/VideoBuffer.h
bcostdolby: I will probably get rid of this and just have a presentation time (no need for DTS either). I wanted to make sure it worked before cleaning it up. I knew in the past we were modifying the clock rates for packetization and started thinking about this change before that work was done.

src/VideoPipe.cpp
bcostdolby: will do

include/VideoBuffer.h
bcostdolby: sounds good

include/rtp/RTPPacket.h
bcostdolby: removed

include/video.h
bcostdolby: Using "timestamp" and presentationTime now only where it makes sense

src/VideoPipe.cpp
bcostdolby: Was already there just updated the code to call it

include/VideoBuffer.h
bcostdolby: done

include/video.h
bcostdolby: Renamed to CopyPresentedTimingInfo() as it is copying only the timing information necessary for a decoded/presented frame

include/video.h
```diff
@@ -131,6 +136,8 @@ class VideoFrame : public MediaFrame
 		ClearCodecConfig();
 		//Clear layers
 		layers.clear();
+
+		presentationTime = 0;
```
murillo128, line 140: should we reset
```
	width		= 0;
	height		= 0;
	targetBitrate	= 0;
	targetFps	= 0;
```

also for completenes?

murillo128 (approved):
lgtm

include/VideoBuffer.h
```diff
@@ -122,15 +122,20 @@ class VideoBuffer
 	DWORD   GetClockRate() const	{ return clockRate.value();		}
 	void    SetClockRate(DWORD clockRate) { this->clockRate = clockRate;	}
 
-
-	void    SetTimingInfo(const VideoBuffer::shared& videoBuffer)
+	void    CopyTimingInfo(const VideoBuffer::const_shared& videoBuffer)
```
harryz2000, line 125: Looks the function name could be CopyTimingInfoIfExists(...).

Otherwise, the current object's value wouldn't be changed. Is it expected?

include/video.h
```diff
@@ -119,6 +121,9 @@ class VideoFrame : public MediaFrame
 	void SetBFrame(bool isBFrame) { this->isBFrame = isBFrame; }
 	bool IsBFrame() const { return isBFrame; }
 
+	void SetPresentationTime(uint64_t v) { presentationTime = v;}
```
harryz2000, line 124: Can we rename the PresetntationTime to be PresentationTimestamp? Also can use use a more meaningful argument name, such as "ts"?

I always think the time is more like a EPOCH time, to similar. While Timestamp would have a special clock rate. 

I feel it would help to understand as we have quite many "time"s.

include/video.h
```diff
@@ -196,13 +204,18 @@ class VideoDecoder
 
 };
 
-inline void CopyTimingInfo(const VideoFrame::const_shared& videoFrame, VideoBuffer::shared& videoBuffer)
+// Used by decoders after potentially reordering so needs to copy all timing info 
+// from the given source to the decoded frame timestamp. 
+//
+// Note: This will copy the presentation time as the reference timestamp now
+// this video data has been decoded/presented
+inline void CopyPresentedTimingInfo(const VideoFrame::const_shared& videoFrame, VideoBuffer::shared& videoBuffer)
```
harryz2000, line 212: It looks it is not only copying the pts, maybe rename it to `CopyFrameTimingInfo`?

include/VideoBuffer.h
murillo128: right, it should reset them before setting them

include/video.h
bcostdolby: The issue with "CopyTimingInfo" is it doesnt indicate it is copying a PTS to a DTS in the name and sounds like it is just copying like VideoBuffer::CopyTimingInfo() that just copys 1:1 field for field.

I.e. This does:
> dts = pts
> ... copy others

The VideoBuffer::CopyTimingInfo is a real copy that does:
> dts = dts
> ... copy others

I.e. This one is changing the meaning of the timestamps at the same time as doing the copy and I dont want that to be missed in reading the code. Its been through a few different names now. Maybe we should change it to do a 1:1 copy, but then the decoders set the dts from the PTS separately?

include/video.h
bcostdolby: done

include/video.h
bcostdolby: done

murillo128 (approved):
lgtm

PR #291, opened 2024-06-06
DIOS-5877 Support transrating in VideoPipe

murillo128:
Completely untested, just to validate the idea

### DIFF
--- a/include/VideoPipe.h
+++ b/include/VideoPipe.h
@@ -68,7 +68,7 @@ private:
 	unsigned int droppedFramesSinceReport = 0;
 	unsigned int totalFramesSinceReport = 0;
 
-	uint32_t maxDelay = 0;
+	uint32_t maxDelayMs = 0;
 };
 
 #endif	/* VIDEOPIPE_H */
--- a/include/VideoPipe.h
+++ b/include/VideoPipe.h
@@ -53,6 +53,7 @@ private:
 	VideoBufferScaler scaler;
 	AllowedDownScaling allowedDownScaling = AllowedDownScaling::Any;
 
+	uint64_t lastGrabbedTimestamp = 0;
 	uint64_t lastDroppedReport = 0;
 	unsigned int droppedFramesSinceReport = 0;
 	unsigned int totalFramesSinceReport = 0;
--- a/include/VideoPipe.h
+++ b/include/VideoPipe.h
@@ -17,6 +17,8 @@ public:
 		SameOrLower	= 1,
 		LowerOnly	= 2
 	};
+	static constexpr uint64_t NoTimestamp = std::numeric_limits<uint64_t>::max();
+
 public:
 	VideoPipe();
 	~VideoPipe() override;
@@ -53,7 +55,7 @@ private:
 	VideoBufferScaler scaler;
 	AllowedDownScaling allowedDownScaling = AllowedDownScaling::Any;
 
-	uint64_t lastGrabbedTimestamp = 0;
+	uint64_t lastGrabbedTimestamp = NoTimestamp;
 	uint64_t lastDroppedReport = 0;
 	unsigned int droppedFramesSinceReport = 0;
 	unsigned int totalFramesSinceReport = 0;
--- a/include/VideoPipe.h
+++ b/include/VideoPipe.h
@@ -11,6 +11,13 @@ class VideoPipe :
 	public VideoInput
 {
 public:
+	// This queue should generally stay mostly empty and only increase in size
+	// if we are falling behind. To place a max limit on the latency use SetMaxDelay()
+	//
+	// Otherwise we will just use this queue size as the max limit. In the usage I see
+	// typically we want to account for at most 1 sec of jitter and usually instead will
+	// limit it to 200msec. 
+	static constexpr size_t MaxOutstandingFramesDefault = 30;
 	enum AllowedDownScaling
 	{
 		Any		= 0,
@@ -25,7 +32,7 @@ public:
 
 	int Init(float scaleResolutionDownBy = 0.0f, uint32_t scaleResolutionToHeight = 0, AllowedDownScaling allowedDownScaling = AllowedDownScaling::Any);
 	int End();
-	void SetMaxDelay(uint32_t maxDelay) { this->maxDelay = maxDelay; };
+	void SetMaxDelay(uint32_t maxDelay);
 
 	/** VideoInput */
 	int StartVideoCapture(uint32_t width, uint32_t height, uint32_t fps) override;
--- a/include/VideoPipe.h
+++ b/include/VideoPipe.h
@@ -36,6 +36,9 @@ public:
 	size_t NextFrame(const VideoBuffer::const_shared& videoBuffer) override;
 	void ClearFrame() override;
 
+
+	void SetMaxDelay(uint32_t maxDelay) { this->maxDelay = maxDelay; };
+
 private:
 	uint32_t videoWidth = 0;
 	uint32_t videoHeight = 0;
@@ -45,7 +48,7 @@ private:
 	int inited = false;
 	int capturing = false;
 	int cancelledGrab = false;
-
+	
 	CircularQueue<VideoBuffer::const_shared> queue;
 
 	pthread_mutex_t newPicMutex;
@@ -59,6 +62,8 @@ private:
 	uint64_t lastDroppedReport = 0;
 	unsigned int droppedFramesSinceReport = 0;
 	unsigned int totalFramesSinceReport = 0;
+
+	uint32_t maxDelay = 0;
 };
 
 #endif	/* VIDEOPIPE_H */
--- a/include/VideoPipe.h
+++ b/include/VideoPipe.h
@@ -25,6 +25,7 @@ public:
 
 	int Init(float scaleResolutionDownBy = 0.0f, uint32_t scaleResolutionToHeight = 0, AllowedDownScaling allowedDownScaling = AllowedDownScaling::Any);
 	int End();
+	void SetMaxDelay(uint32_t maxDelay) { this->maxDelay = maxDelay; };
 
 	/** VideoInput */
 	int StartVideoCapture(uint32_t width, uint32_t height, uint32_t fps) override;
@@ -35,9 +36,6 @@ public:
 	/** VideoOutput */
 	size_t NextFrame(const VideoBuffer::const_shared& videoBuffer) override;
 	void ClearFrame() override;
-
-
-	void SetMaxDelay(uint32_t maxDelay) { this->maxDelay = maxDelay; };
 
 private:
 	uint32_t videoWidth = 0;
--- a/src/VideoEncoderWorker.cpp
+++ b/src/VideoEncoderWorker.cpp
@@ -174,9 +174,6 @@ int VideoEncoderWorker::Encode()
 	//Set bitrate
 	videoEncoder->SetFrameRate(fps,bitrate,intraPeriod);
 
-	//No wait for first
-	QWORD frameTime = 0;
-
 	//Iniciamos el tamama�o del encoder
  	videoEncoder->SetSize(width,height);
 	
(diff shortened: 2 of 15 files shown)


src/VideoPipe.cpp
```diff
@@ -134,60 +134,64 @@ VideoBuffer::const_shared VideoPipe::GrabFrame(uint32_t timeout)
...
-		queue.pop_front();
-	}
+		//Get current image if there is anything in the queue
+		if (!queue.empty())
+		{
+			videoBuffer = queue.front();
+			queue.pop_front();
+		}
+	//Ignore all the frames before the next timestamp to match the capture fps
+	} while (videoBuffer && videoBuffer->HasTimestamp() && lastTimestamp + videoBuffer->GetClockRate()/videoFPS >= videoBuffer->GetTimestamp());
```
bcostdolby, line 194: Should the >= just be a > ?

I.e. Do we want to drop this frame if it matches the next expected?

Also a little worried we might drop things we dont need to if timestamps are a bit off.

murillo128: Yes, that should be a >=

src/VideoEncoderWorker.cpp
```diff
@@ -241,7 +241,7 @@ int VideoEncoderWorker::Encode()
 		fpsAcu.Update(getTime()/1000,1);
 
 		//Check
-		if (frameTime)
+	/*	if (frameTime)
```
bcostdolby, line 244: i figure we are just removing this code and relevant support members/variables?

murillo128: I think so yes. Although we would be makeng the timestamp field of the videoBuffer mandatory to work with the encoder, which I think I am fine tine, but we would have to double check

murillo128 (commented):
lgtm

src/VideoPipe.cpp
```diff
@@ -134,60 +134,77 @@ VideoBuffer::const_shared VideoPipe::GrabFrame(uint32_t timeout)
...
+		{
+			videoBuffer = queue.front();
+			queue.pop_front();
+		}
+	//Ignore all the frames before the next timestamp to match the capture fps
+	} while (videoBuffer 
+		&& videoFPS
+		&& lastGrabbedTimestamp!= NoTimestamp 
+		&& videoBuffer->HasTimestamp() 
+		&& lastGrabbedTimestamp + videoBuffer->GetClockRate()/videoFPS - 1 > videoBuffer->GetTimestamp()
```
bcostdolby, line 206: Still need to think this last line of the condition through a bit more. 

I think there is the chance we can have a slightly different FPS than we want with this. Instead of keeping track of OUR encoders timestamp we are using last pulled + diff. 

If last pulled is off by a small amount as can happen for quantization, then we might keep re-syncing our anchor slightly. 

Should I change this to keep track of our timestamp (first frame set it) like in lastGrabbedTimestamp, and then just increment this by the difference instead of reset to last grabbed again?

test/unit/TestVideoPipe.cpp
```diff
@@ -289,3 +289,280 @@ TEST(TestVideoPipe, pictureResizeScaleDownBy)
...
+        for (int i = 0; i < 30; ++i)
+        {
+                VideoBuffer::shared sharedVidBuffer = std::make_shared<VideoBuffer>(width, height);
+                sharedVidBuffer->SetClockRate(clockrate);
+                sharedVidBuffer->SetTimestamp(i * (double)clockrate / infps);
+                auto queued = vidPipe.NextFrame(sharedVidBuffer);
+                ASSERT_EQ(queued, i + 1);
+        }
+
+        for (int i = 0; i < 15; ++i)
```
yezhan10, line 414: is 15 a typo? should it be 30 instead?

src/VideoPipe.cpp
```diff
@@ -279,10 +305,25 @@ size_t VideoPipe::NextFrame(const VideoBuffer::const_shared& videoBuffer)
 
 	auto now = getTime();
 
+
+	//Do not enqueue more than the max delay
+	while (maxDelay
```
harryz2000, line 310: Personally I would prefer no to include many conditions in the while(), especially some conditions are not changing at all so it wouldn't be efficient. I found it is not easy to check. Maybe it could be changed to something similar as below (not the code was not tested! Just for showing my point):

```
	if (maxDelay && videoBuffer->HasTimestamp())
	{
		auto vbClockRate = videoBuffer->GetClockRate();
		auto vbTimestamp = videoBuffer->GetTimestamp();
		
		auto maxDelayInClockRate = (maxDelay * vbClockRate) / 1000;
		
		//Do not enqueue more than the max delay
		while (!queue.empty())
		{
			auto& front = queue.front();
			
			if (!front->HasTimestamp()) break;
			
			if (vbClockRate != front->GetClockRate()) break;
			if (vbTimestamp < front->GetTimestamp() break;
			
			if (vbTimestamp - front->GetTimestamp() < maxDelayInClockRate)
			{
				break;
			}
			
			//Drop first picture
			queue.pop_front();
			++droppedFramesSinceReport;
		}
	}

```

test/unit/TestVideoPipe.cpp
yezhan10: just noticed that downrating and downratingNotExact use different clockrate, why we dont need upratingNotExact?

test/unit/TestVideoPipe.cpp
```diff
@@ -289,3 +289,280 @@ TEST(TestVideoPipe, pictureResizeScaleDownBy)
...
+        int clockrate = 90000;
+
+        int width = 640;
+        int height = 480;
+
+        VideoPipe vidPipe;
+        vidPipe.Init();
+        vidPipe.StartVideoCapture(width, height, outfps);
+
+        //Enqueue 10 frames
```
harryz2000, line 307: typo? 10 or 30?

harryz2000: There are similar typos in other places as well.

src/VideoPipe.cpp
```diff
@@ -279,10 +305,25 @@ size_t VideoPipe::NextFrame(const VideoBuffer::const_shared& videoBuffer)
...
 
+
+	//Do not enqueue more than the max delay
+	while (maxDelay 
+		&& !queue.empty() 
+		&& videoBuffer->HasTimestamp() 
+		&& queue.front()->HasTimestamp()
+		&& videoBuffer->GetTimestamp() > queue.front()->GetTimestamp()
+		&& videoBuffer->GetClockRate() == queue.front()->GetClockRate()
+		&& (videoBuffer->GetTimestamp() - queue.front()->GetTimestamp()) >= (maxDelay * videoBuffer->GetClockRate()) / 1000
```
yezhan10, line 316: according to the code, the maxDelay is in time, curious why it can not be in frame count?

test/unit/TestVideoPipe.cpp
```diff
@@ -289,3 +289,280 @@ TEST(TestVideoPipe, pictureResizeScaleDownBy)
...
+        }
+
+        // Start grab frame that is waiting for the next input
+        // for at most 10msec, however with bug will wait forever
+        auto blocking_grab_frame = std::async(std::launch::async, [&]()->auto{
+            return vidPipe.GrabFrame(10);
+        });
+
+        auto pic = blocking_grab_frame.get();
+        ASSERT_FALSE(pic);
```
harryz2000, line 567: I think it would be good to use
```suggestion
        ASSERT_EQ(nullptr, pic);
```

src/VideoPipe.cpp
harryz2000: Looks OK to me. Just as another comment, my personal preference would be not include many conditions in while() especially for those not changing conditions. Having said that, this one only have below variables not changing.

```
		&& videoFPS
		&& lastGrabbedTimestamp!= NoTimestamp 
```

harryz2000 (commented):
Just a few minor suggestions.

test/unit/TestVideoPipe.cpp
bcostdolby: ANY type of uprating behaves the same as ifps == ofps. I.e. We never actually uprate. Exact or not exact hit the same condition. I can add a test to verify this. It makes sense.

test/unit/TestVideoPipe.cpp
bcostdolby: yeah I am just removing the comments. they dont add anything useful

src/VideoPipe.cpp
bcostdolby: max delay in other parts of the system related to this is in msec. This will be used by the JS layer to set bounds on how much delay we want. If it is in frames then those parts of the code will need to know the fps which they dont currently have.

reminds me though I should probably update those parts of the system and test with it properly.

test/unit/TestVideoPipe.cpp
bcostdolby: done in all places and also inverse for ASSERT_TRUE(pic)

src/VideoPipe.cpp
bcostdolby: I thought about moving them out before as well after seeing Sergio' original change, but they are very relevant to the remainder of the condition.

I.e. We are making sure the remainder of the conditional (the main part) is actually valid to compute. 

Also because it is a do-while not a while moving them out is more annoying.

If ok, I would like to leave this one as is but add some comments

harryz2000: Feel free to leave it as it is if you prefer. I don't think more comments is needed.

bcostdolby:
Based on the review from Ying, I also updated the SetMaxDelay() to check if it needs to increase the size of the queue and wrote a test for this. I think it is worth re-reviewing this change @harryz2000 and @yezhan10

murillo128 (commented):
lgtm

src/VideoPipe.cpp
```diff
@@ -63,6 +62,30 @@ int VideoPipe::End()
...
+	// Note: We dont *know* the ingest fps, so will just assume max is 60fps for now
+	// If for whatever reason the egress fps is larger we will know about it here
+	int maxFPS = std::max(60, videoFPS);
+
+	// Figure out the max frames we might every have in this queue. 
+	// Note: Actual latency values are enforced by the maxDelay
+	// we are providing an upper limit on the outstanding items in the queue
+	// and will resize this queue to be larger in case we increase the latency
+	// past what is likely to be supported by the queues default
+	size_t maxDelayInFrames = std::ceil((maxDelay * maxFPS) / 1000.0);
```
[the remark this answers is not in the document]
bcostdolby: There are some super weird edge/error cases where the queue could become unbounded so chose not to, but happy to change to use it.

Some examples are no timestamps or always changing clock rates or timestamps going backwards or SetMaxDelay not set.

These are not really valid scenarios but the max size protects against them (potentially from a malicious actor as well).

The other thing was that we currently have SetMaxDelay support a value of 0 (which isnt used) which means use the queue max I  was thinking to use auto grow I would probably want to change queue to support changing auto-grow or not based on if the delay is 0. We could also set the delay to some sensible default and not permit a value of 0?

Any preferences?

include/VideoPipe.h
```diff
@@ -53,9 +63,12 @@ class VideoPipe :
 	VideoBufferScaler scaler;
 	AllowedDownScaling allowedDownScaling = AllowedDownScaling::Any;
 
+	uint64_t lastGrabbedTimestamp = NoTimestamp;
 	uint64_t lastDroppedReport = 0;
 	unsigned int droppedFramesSinceReport = 0;
 	unsigned int totalFramesSinceReport = 0;
+
+	uint32_t maxDelay = 0;
```
harryz2000, line 71: Consider add unit to the variable name (preferrable), i.e. maxDelayMs, or add a comment.

bcostdolby: done

bcostdolby:
@harryz2000 please approve if you are happy with it now

src/VideoPipe.cpp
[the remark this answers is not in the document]
murillo128: how about limiting the autoGrow max size on the queue? https://github.com/medooze/media-server/pull/295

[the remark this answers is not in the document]
murillo128: umm.. I don't think my modification of the queue fixes much.. your change looks good to me

yezhan10 (approved):

harryz2000 (approved):

PR #293, opened 2024-06-13
Report RTMP client error for event logging

harryz2000:


src/rtmp/rtmpclientconnection.cpp
```diff
@@ -82,8 +82,15 @@ int RTMPClientConnection::Connect(const char* server, int port, const char* app,
 
 	//If not found
 	if (!host)
+	{
 		//Error
+		if (listener)
+		{
+			listener->onError(this, RTMPClientConnection::ErrorCode::FailedToResolveURL);
```
murillo128, line 89: Given that the Connect is a pure synchronous method, I would prefer to return the error code instead of firing an error event in this case.

src/rtmp/rtmpclientconnection.cpp
```diff
@@ -97,8 +104,15 @@ int RTMPClientConnection::Connect(const char* server, int port, const char* app,
 // Ignore coverity error: "this->fd" is passed to a parameter that cannot be negative.
 // coverity[negative_returns]
 	if (connect(fd, (sockaddr*)&addr, sizeof(addr)) < 0 && errno != EINPROGRESS)
+	{
+		if (listener)
+		{
+			listener->onError(this, RTMPClientConnection::ErrorCode::FailedToConnectSocket);
```
murillo128, line 110: same here

src/rtmp/rtmpclientconnection.cpp
```diff
@@ -306,6 +329,9 @@ int RTMPClientConnection::Run()
 				Error("-RTMPClientConnection::Run() Exception parsing data: %s\n", e.what());
 				//Dump it
 				Dump(data, len);
+				
+				if (listener) listener->onError(this, RTMPClientConnection::ErrorCode::FailedToParseData);
```
murillo128, line 333: I don't think a parsing error should trigger an error, it could be harmless, no need to close the connection due to that.

murillo128 (changes requested):

src/rtmp/rtmpclientconnection.cpp
harryz2000: Yes. I thought of that. In that case, we would need to change the return value to be something like 0 means success. Will look into it.

src/rtmp/rtmpclientconnection.cpp
harryz2000: Sound reasonable. The existing code actually quits the thread. If we regard it harmless, I would change it to be "continue" instead.

murillo128: can you check what we do on the server connections?

src/rtmp/rtmpclientconnection.cpp
murillo128: yes, maybe returning an enum could be a good idea. @mildsunrise wdyt?

src/rtmp/rtmpclientconnection.cpp
harryz2000: The server connection is like current implementation. It will "break" the loop, which means it would quit the thread in case of parse error. https://github.com/medooze/media-server/blob/c20ff2202258e03713ddb309a1da29bc5ba084e8/src/rtmp/rtmpconnection.cpp#L244

murillo128 (approved):
lgtm

PR #294, opened 2024-06-17
DIOS-5789: add unit test for bframe

yezhan10:
1. add bframe unit tests for all supported decoders(h264/h265/vp8/vp9)
2. add flushing logic in h265decoder 
3. copy timing info in vp9

### DIFF
--- a/test/unit/TestBFrame.cpp
+++ b/test/unit/TestBFrame.cpp
@@ -1,5 +1,6 @@
 #include "TestCommon.h"
-extern "C" {
+extern "C"
+{
 #include <libswscale/swscale.h>
 #include <libavcodec/avcodec.h>
 #include <libavformat/avformat.h>
@@ -8,7 +9,8 @@ extern "C" {
 #include "VideoDecoderWorker.h"
 #include "VideoPipe.h"
 
-struct EncodingParams {
+struct EncodingParams
+{
     int width;
     int height;
     int fps;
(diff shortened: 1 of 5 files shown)


test/unit/TestBFrame.cpp
```diff
@@ -0,0 +1,302 @@
...
+            /* Y */
+            for (y = 0; y < height; y++)
+            {
+                for (x = 0; x < width; x++)
+                {
+                    frame->data[0][y * frame->linesize[0] + x] = i * 10 + 1;
+                }
+            }
+            /* Cb and Cr */
+            for (y = 0; y < height / 2; y++)
```
bcostdolby, line 59: Maybe just add a comment describing what sort of image this is generating and where this came from.

I assume it is just covering the colour space, but I dont understand it (not without reading some external docs)

Also this is obviously truncating the integers. I assume this will produce some compile warnings we should probably clean up.

test/unit/TestBFrame.cpp
```diff
@@ -0,0 +1,302 @@
...
+    {
+        for (int i = 0; i < numPackets; ++i)
+        {
+            int x, y;
+            /* Y */
+            for (y = 0; y < height; y++)
+            {
+                for (x = 0; x < width; x++)
+                {
+                    frame->data[0][y * frame->linesize[0] + x] = i * 10 + 1;
```
bcostdolby, line 55: Be worth adding this into a function that can be used below to provide the expected values.

test/unit/TestBFrame.cpp
```diff
@@ -0,0 +1,302 @@
...
+    std::vector<uint8_t> generatedPixelValue;
+    std::vector<uint8_t> expectedPixelValue;
+
+    std::vector<int> generatedPTS;
+    std::vector<int> expectedPTS;
+
+    for (int i = 0; i < numPackets; ++i)
+    {
+        // same equation used in generateAVPackets function
+        expectedPixelValue.push_back(10 * i + 1);
```
bcostdolby, line 188: Just create a function that we can use in both places.

bcostdolby (changes requested):

test/unit/TestBFrame.cpp
```diff
@@ -0,0 +1,312 @@
...
+        EXPECT_EQ(expectedPTS[i], generatedPTS[i]) << "pts differ at index " << i;
+    }
+
+    for (int i = 0; i < expectedPixelValue.size(); ++i)
+    {
+        EXPECT_EQ(expectedPixelValue[i], generatedPixelValue[i]) << "pixel value differ at index " << i;
+    }
+}
+
+TEST(TestBFrame, H264CodecBFrame)
```
phil-moss-dolby, line 255: Since we're moving in the direction of having 1:1 source file to test file structure, should we keep this TestBFrame.cpp as generic test code, but have the tests themselves running in their appropriate source files? h264 test belongs in the Testh264decoder.cpp, and that test calls runBFrameTest defining its own codec and other params. Just a file structure and organizational question

bcostdolby (approved):

harryz2000 (approved):
lgtm

PR #296, opened 2024-06-21
feat [DIOS-5853] Media server - Handle scte35 messages as unregistered SEI user data

amaha877:


include/MediaFrameListenerBridge.h
```diff
@@ -68,6 +68,10 @@ class MediaFrameListenerBridge :
 	virtual int SendPLI(DWORD ssrc) override { return 1; };
 	virtual int Reset(DWORD ssrc) override { return 1; };
 
+	void SetScteData(Buffer data);
```
murillo128, line 71: do not use the MediaFrameListenerBridge as intermediary

src/h264/H264Packetizer.h
```diff
@@ -16,6 +18,11 @@ class H264Packetizer : public H26xPacketizer
 	H264Packetizer();
 	std::unique_ptr<MediaFrame> ProcessAU(BufferReader &reader) override;
 
+	void SetScteData(Buffer data);
+	std::optional<Buffer> GetScteData();
```
murillo128, line 22: no need for a getter

src/h264/H264Packetizer.cpp
```diff
@@ -175,6 +178,85 @@ void H264Packetizer::OnNal(VideoFrame& frame, BufferReader& reader, std::optiona
...
+		sei.push_back(0x06);	// SEI NAL
+		sei.push_back(0x05);	// User unregistered data
+		
+		uint8_t uuid[16] = {
+			0x2b, 0x69, 0xba, 0x1b, 0x77, 0x7e, 0x46, 0x53,
+			0x99, 0x7a, 0xc6, 0x75, 0xb9, 0xd3, 0xac, 0xaf
+		};
+		sei.insert(sei.end(), uuid, uuid + sizeof(uuid));
+		
+		/* Construct json encoded as a string
```
murillo128, line 195: please use a binary format instead

src/h264/H264Packetizer.cpp
```diff
@@ -175,6 +178,85 @@ void H264Packetizer::OnNal(VideoFrame& frame, BufferReader& reader, std::optiona
...
+		Example:
+		{
+  			id: 14194058,
+  			start: 1717005742629,
+  			duration: 30,
+	  		tag: '/DBcAAAAAAAAAP/wBQb//ciI8QBGAh1DVUVJXQk9EX+fAQ5FUDAxODAzODQwMDY2NiEEZAIZQ1VFSV0JPRF/3wABLit7AQVDMTQ2NDABAQEKQ1VFSQCAMTUwKnPhdcU='
+		}
+		 We need to compute a unique hash for every scte payload received.
```
murillo128, line 203: why? a unique has based on the scte data provides no info, if you need an unique id, use a counter instead

src/h264/H264Packetizer.cpp
```diff
@@ -175,6 +178,85 @@ void H264Packetizer::OnNal(VideoFrame& frame, BufferReader& reader, std::optiona
...
+			0x99, 0x7a, 0xc6, 0x75, 0xb9, 0xd3, 0xac, 0xaf
+		};
+		sei.insert(sei.end(), uuid, uuid + sizeof(uuid));
+		
+		/* Construct json encoded as a string
+		Example:
+		{
+  			id: 14194058,
+  			start: 1717005742629,
+  			duration: 30,
```
murillo128, line 200: duration is not used, please remove

src/h264/H264Packetizer.h
```diff
@@ -16,6 +18,11 @@ class H264Packetizer : public H26xPacketizer
 	H264Packetizer();
 	std::unique_ptr<MediaFrame> ProcessAU(BufferReader &reader) override;
 
+	void SetScteData(Buffer data);
```
murillo128, line 21: this is not really a setter, use a single method:
```suggestion
	void AppendScteData(uint64_t ts, const Buffer& data, uint8_t repeatCount);
```

src/h264/H264Packetizer.cpp
```diff
@@ -3,10 +3,13 @@
 #include "video.h"
 #include "h264/h264.h"
 
+constexpr int FrameRepeatCount = 20;
```
murillo128, line 6: better to have this configurable from outside (see below)

src/h264/H264Packetizer.cpp
```diff
@@ -175,6 +178,85 @@ void H264Packetizer::OnNal(VideoFrame& frame, BufferReader& reader, std::optiona
 		frame.AllocateCodecConfig(config.GetSize());
 		config.Serialize(frame.GetCodecConfigData(), frame.GetCodecConfigSize());
 	}
+	
+	if (scteMessage.GetSize() != 0)
+	{	
+		//Add unregistered SEI message NAL
+		std::vector<uint8_t> sei;
```
murillo128, line 185: Move the serialization to the `AppendScteData` function, so we do it only once

src/h264/H264Packetizer.h
```diff
@@ -4,6 +4,8 @@
 #include "H26xPacketizer.h"
 #include "avcdescriptor.h"
 
+#include <queue>
```
murillo128, line 7: not needed, right?

murillo128 (changes requested):
please remove the mp4v2 from the PR

include/MediaFrameListenerBridge.h
harryz2000: Yeah. Agree. This is not ideal, probably only works for prototyping. As this involves passing a data from one stream to another stream, to properly passing, we might have to find a way to pass data across different streams, then to H264Packetizer.

amaha877: Ok this was the part i wasn't too sure on how to refactor. Sounds like adding a scte listener directly to MpegTSAVCStream and MpegTSHEVCStream could be an option per harry's suggestion. I will give that a go.

src/h264/H264Packetizer.cpp
amaha877: I assumed that was the requirement but yes a counter would do just fine in this case.

src/h264/H264Packetizer.h
amaha877: ok

src/h264/H264Packetizer.cpp
amaha877: Agreed that transmitting raw json as a stream of bytes ins't the best. Are you using something like protobuf / some other binary encoding format elsewhere in the project?

If not, is it worth designing our own for ~100-200 bytes of data every now and then

murillo128: protobuf is an overkill, 2 bytes for counter,  8 bytes for timestamp and the rest of the payload the scte message

include/MediaFrameListenerBridge.h
amaha877: Now using a SCTEPublisher and SCTEReceiver interface for notifying streams directly. Thanks Harry for the suggestion

src/h264/H264Packetizer.cpp
amaha877: Done and discussed the same with SDK team

include/MediaFrameListenerBridge.h
```diff
@@ -68,15 +68,14 @@ class MediaFrameListenerBridge :
...
 
-
 private:
 	void Dispatch(const std::vector<RTPPacket::shared>& packet);
         
 public:
 	TimeService& timeService;
 	Timer::shared dispatchTimer;
 
-	std::queue<PacketScheduleInfo> packets;
```
murillo128, line 79: remove extra changes

include/tools.h
```diff
@@ -513,4 +513,5 @@ inline size_t CountBits(uint64_t val)
 	}
 	return count;
 }
+
```
murillo128, line 516: remove extra changes

src/h264/H264Packetizer.h
```diff
@@ -24,6 +26,17 @@ class H264Packetizer : public H26xPacketizer
...
+	
+	Buffer scteMessage;
+	uint64_t scteTimestamp;
+	// To account for potential packet loss, we repeat the SEI containing 
+	// scte tags for 20 frames to provide added resiliency.
+	// Also, we assume that another scte message cannot arrive within is interval
+	uint8_t scteFrameRepeatCount;
+	uint8_t repeatFrameCounter;
+	//Unregistered SEI message NAL
+	std::vector<uint8_t> sei;
```
murillo128, line 38: better use Buffer
```suggestion
	Buffer sei;
```

src/h264/H264Packetizer.cpp
```diff
@@ -186,3 +211,50 @@ std::unique_ptr<MediaFrame> H264Packetizer::ProcessAU(BufferReader& reader)
 
 	return H26xPacketizer::ProcessAU(reader);
 }
+
+void H264Packetizer::AppendScteData(uint64_t ts, const Buffer& data, uint8_t repeatCount)
+{
+	scteMessage.AppendData(data.GetData(), data.GetSize());
```
murillo128, line 217: why are we storing the `scteMessage`?

src/h264/H264Packetizer.cpp
```diff
@@ -186,3 +211,50 @@ std::unique_ptr<MediaFrame> H264Packetizer::ProcessAU(BufferReader& reader)
...
+	scteFrameRepeatCount = repeatFrameCounter = repeatCount;
+	scteTimestamp = ts;
+
+	// Clear just the scte payload and retain the 2 byte code, size and 16 bytes UUID
+	sei.erase(sei.begin() + 2 + 1 + 16, sei.end());
+
+	// Binary format - SEI NAL (0x06) + Unregistered User data code (0x05) + UUID (16 bytes) + SCTE ID (2 bytes) + TS (8 bytes) + Base64 encoded scte payload (x bytes)
+	
+	// Convert scte data to base64
+	uint8_t scteBase64[scteMessage.GetSize()*2] = {0};
```
murillo128, line 227: there is no need to convert it to base64 here, we can send the raw binary bytes to reduce the total size

src/h264/H264Packetizer.cpp
```diff
@@ -186,3 +211,50 @@ std::unique_ptr<MediaFrame> H264Packetizer::ProcessAU(BufferReader& reader)
...
+		{
+			base64Len = i;
+			break;
+		}
+	}
+
+	// Modify the payload size at pos 2 (zero indexed) 
+	// [0] 0x06 SEI NAL 
+	// [1] 0x05 Unregistered user data
+	// [2] Payload size
```
murillo128, line 243: we don't really need the payload size, it will be the rest of the NAL data

src/h264/H264Packetizer.cpp
```diff
@@ -175,6 +184,22 @@ void H264Packetizer::OnNal(VideoFrame& frame, BufferReader& reader, std::optiona
 		frame.AllocateCodecConfig(config.GetSize());
 		config.Serialize(frame.GetCodecConfigData(), frame.GetCodecConfigSize());
 	}
+	
+	if (scteMessage.GetSize() != 0)
+	{	
+		//Escape nal
+		uint8_t seiEscaped[sei.size()*2];
+		auto seiSize = NalEscapeRbsp(seiEscaped, sizeof(seiEscaped), sei.data(), sei.size()).value();
```
murillo128, line 192: `better scape it on the `AppendScteData()` if using buffers, this functions should be:


```
if (repeatFrameCounter>0 && !sei.IsEmpty())
{
  EmitNal(frame, BufferReader reader(sei));
  repeatFrameCounter--;
}

File: src/h264/H264Packetizer.cpp [hunk 17]
```diff
@@ -3,10 +3,19 @@
...
-
 H264Packetizer::H264Packetizer() : H26xPacketizer(VideoCodec::H264)
 {
-
+	sei.clear();            // Flush any previous data
+	sei.push_back(0x06);	// SEI NAL
+	sei.push_back(0x05);	// User unregistered data
+	sei.push_back(0x00);    // Payload size
+	
+	uint8_t uuid[16] = {
```
murillo128, line 13:
```suggestion
	constexpr std::array<uint8_t,16> uuid = {
```

src/h264/H264Packetizer.cpp
```diff
@@ -186,3 +211,50 @@ std::unique_ptr<MediaFrame> H264Packetizer::ProcessAU(BufferReader& reader)
...
+	}
+
+	// Modify the payload size at pos 2 (zero indexed) 
+	// [0] 0x06 SEI NAL 
+	// [1] 0x05 Unregistered user data
+	// [2] Payload size
+	sei.at(2) = 16 + 2 + 8 + base64Len; // UUID 16 bytes + SCTE ID (2 bytes) + TS (8 bytes) + Base64 encoded scte payload (x bytes)
+
+	// Insert 2 byte scte message ID
+	scteMessageId++;
```
murillo128, line 247: This could be simplified by

```
sei.Alloc(16+2+data.Size()*2)
BufferWritter(sei);
writter.Set(uuid);
writter.Set2(scteMessageId++);
auto pos = writter.Set8(timestamp);
auto size = NalEscapeRbsp(sei.GetData() + pos , writter.GetLeft(), data.GetData(), data.GetSize());
writter.Skip(size);
sei.SetSize(writter.GetLength());
```

We should add a 
```
 uint8_t GetData() { return data + pos: }
```
to the https://github.com/medooze/libdatachannels/blob/8a52932f80f7fdb569d2ff56f0481efcfe4a405f/src/internal/BufferWritter.h

so we could a NalEscape function based on writters/buffers

```
bool NalEscapeRbsp(BufferWritter& writter, const Buffer& buffer)
{
  auto size =  NalEscapeRbsp(writter.GetData() , writter.GetLeft(), data.GetData(), data.GetSize());
  writter.Skip(size);
  return size != 0;
}
```

and simplify the function even more to:
```
sei.Alloc(16+2+data.Size()*2)
BufferWritter(sei);
writter.Set(uuid);
writter.Set2(scteMessageId++);
writter.Set8(timestamp);
NalEscapeRbsp(writter, data);
sei.SetSize(writter.GetLength());
```

murillo128 (changes requested):

src/h264/H264Packetizer.cpp
amaha877: Because we need to reuse the same scteMessage to provide inter frame redundancy. The appendscte method will only be called when we have an event.

src/h264/H264Packetizer.cpp
amaha877: Regarding base64, that was requested by the Client team. I obliged since it was a few extra bytes and we had a routine handy

src/h264/H264Packetizer.cpp
amaha877: Oh I see your point actually, we can just directly append it to the final SEI message.

src/h264/H264Packetizer.cpp
murillo128: let's discuss that internally

murillo128: re, payloadSize, note that it is an input tobthe serialization function but it is not serialized

src/h264/H264Packetizer.cpp
murillo128: my bad, we need to scape the whole mal unit not just the scte data, the function would be even easier

Buffer unscaped(16+2+data.Size())
BufferWritter(unscaped);
writter.Set(uuid);
writter.Set2(scteMessageId++);
writter.Set8(timestamp);
writter.Set(data);
sei.SetSize(writter.GetLength());


and then scape it later as you already have

src/h264/H264Packetizer.cpp
phil-moss-dolby: Ticket created to finalize payload format (let's make sure client team is involved in this discussion to prevent the back and forth, every time we update they have to as well): https://jira.dolby.net/jira/browse/DIOS-6124 

In the meantime, let's address the functional comments and get this merged to unblock the client team as they would like to start demoing next week. Unless if we feel we can agree on something before Monday, I don't believe the final formatting should block this functionality while @murillo128 is on vacation

phil-moss-dolby:
Merging to unblock Bally's testing and demos in stage, quick follow up: https://jira.dolby.net/jira/browse/DIOS-6124. If there are side effects during QA testing this will be removed from the build

PR #297, opened 2024-06-21
Support RTMPS client

harryz2000:


murillo128:
Please use CamelCase for method names

murillo128:
@mildsunrise ptal

src/rtmp/rtmpclientconnection.cpp
```diff
@@ -407,18 +444,64 @@ DWORD RTMPClientConnection::SerializeChunkData(BYTE* data, DWORD size)
...
+			if (listener) listener->onDisconnected(this, RTMPClientConnection::ErrorCode::TlsEncryptError);
+			return;
+		}
+	}
+	else
+	{
+		WriteData(data, size);
+	}
+	
+	outBytes += size;
```
murillo128, line 464: should we do it inside the `WriteData` instead?

murillo128 (commented):
have not checked the specific code for TLS, the non-tls related changes seems good to me, although I would prefer a class abstracting the client connection and don't have to do `if (tls)`

include/TlsClient.h
```diff
@@ -0,0 +1,100 @@
...
+	
+	void QueueEncryptedData(const uint8_t* data, size_t size);
+	
+	void QueueDecryptedData(const uint8_t* data, size_t size);
+
+	SSL_CTX *ctx = nullptr;
+	SSL *ssl = nullptr;
+
+	BIO *rbio = nullptr;
+	BIO *wbio = nullptr;
```
mildsunrise, line 88: based on my experiences with OpenSSL, I'd strongly recommend wrapping these in `std::unique_ptr` with custom deleters... I'd even suggest copying (some of) these into our repo: https://github.com/nodejs/node/blob/b9289a6e29e54beeaa3f781bde1195e48df0da75/src/crypto/crypto_util.h#L57-L80

the definition of `DeleteFnPtr` can be found here: https://github.com/nodejs/node/blob/b9289a6e29e54beeaa3f781bde1195e48df0da75/src/util.h#L686-L693

src/TlsClient.cpp
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	// Set client mode
+	SSL_set_connect_state(ssl);
+
+	SSL_set_bio(ssl, rbio, wbio);
+	
+	if (hostname)
+	{
+		SSL_set_tlsext_host_name(ssl, hostname);
```
mildsunrise, line 86: I may be missing something (and my OpenSSL knowledge is rusty) but I think this isn't actually verifying the hostname of the certificate, is it? someone could MITM the connection with a certificate that is valid, but for another host. I think you should call [`SSL_set_hostflags` and `SSL_set1_host`](https://www.openssl.org/docs/man3.0/man3/SSL_set1_host.html) as indicated [here](https://wiki.openssl.org/index.php/Hostname_validation). @murillo128 I don't think we need to support TLS connections to bare IPs, right?

mildsunrise (approved):
lgtm except for the security issue... but again, my OpenSSL knowledge is rusty

mildsunrise:
also @harryz2000 sorry for the late review, I'm very bad at organizing my time :_

src/rtmp/rtmpclientconnection.cpp
harryz2000: If we put it to WriteData(), it would be the size of encrypted data, which would be different to the unencrypted data. I would change that. That sounds reasonable.

harryz2000:
> have not checked the specific code for TLS, the non-tls related changes seems good to me, although I would prefer a class abstracting the client connection and don't have to do `if (tls)`

That would involve more change, including need to create different instances in the node js code. But that was my original thought too. I would rethink about that.

include/TlsClient.h
harryz2000: Yep. Use smart pointer would be safer.

src/TlsClient.cpp
harryz2000: No. It is for "[Server Name Indication](https://knowledge.digicert.com/quovadis/ssl-certificates/ssl-general-topics/what-is-sni-server-name-indication)". 

This is possibly not necessary as our server mostly likely only has one hostname. Just put there in case.

harryz2000:
Created another class RTMPSClientConnection for rtmps

src/rtmp/rtmpclientconnection.cpp
```diff
@@ -407,18 +409,29 @@ DWORD RTMPClientConnection::SerializeChunkData(BYTE* data, DWORD size)
 	return len;
 }
 
+void RTMPClientConnection::addPendingRtmpData(const uint8_t* data, size_t size)
```
murillo128, line 412: please use CamelCase for c++ method names

src/rtmp/rtmpclientconnection.cpp
```diff
@@ -240,55 +240,58 @@ int RTMPClientConnection::Run()
...
 			}
-
-			//Write data buffer
-			DWORD len = SerializeChunkData(data, size);
-			//Send it
-			WriteData(data, len);
-			//Increase sent bytes
-			outBytes += len;
+			
+			OnReadyToTransfer();
```
murillo128, line 294: why do we need to signal `OnReadyToTransfer()` ?

murillo128 (commented):
can't we avoid the queuing/dequeing of memory in TlsClient? it is going to kill performance.

src/rtmp/rtmpclientconnection.cpp
harryz2000: OnReadyToTransfer() is to transfer the encrypted tls data. Some data maybe generated by openssl internally before connected, i.e the handshake data. Not matter whether we have RTMP data to send, we need check whether there is TLS data to send.

harryz2000:
> can't we avoid the queuing/dequeing of memory in TlsClient? it is going to kill performance.

I think we would need some sort of buffer to store the data. I will see if I can use a circular buffer. In case of the buffer is full, we may have to drop chunks, which should rarely happen we normally pop them immediately.

harryz2000:
I changed to use callbacks instead of queue for performance.

include/TlsClient.h
```diff
@@ -0,0 +1,90 @@
...
+struct FunctionDeleter {
+  void operator()(T* pointer) const { function(pointer); }
+  typedef std::unique_ptr<T, FunctionDeleter> Pointer;
+};
+
+template <typename T, void (*function)(T*)>
+using DeleteFnPtr = typename FunctionDeleter<T, function>::Pointer;
+
+using SSLCtxPointer = DeleteFnPtr<SSL_CTX, SSL_CTX_free>;
+using SSLPointer = DeleteFnPtr<SSL, SSL_free>;
```
murillo128, line 23: why can't we just use the std::unique_ptr without all this templates?

murillo128 (approved):
lgtm, only minor comments

include/TlsClient.h
harryz2000: We could. Those are something from the page Alba mentioned.

harryz2000: Actually, using these templates would be convenient for unique_ptr, which doesn't allow a deleter to be passed to the constructor. It is different with shared_ptr. I don't think we should use shared_ptr here. So if we don't use the templates, we need to declare functors to make it call the *_free function.

murillo128: Can we move the deleters to OpenSSL.h
```
template <typename T, void (*function)(T*)>
struct SSLFunctionDeleter {
  void operator()(T* pointer) const { function(pointer); }
};

using SSLCtxDeleter = SSLFunctionDeleter<SSL_CTX, SSL_CTX_free>;
using SSLDeleter = SSLFunctionDeleter<SSL, SSL_free>;
```

and use a std::unique_ptr in the class directly?

```
 std::unique_ptr<SSL, SSLDeleter> ssl;
 std::unique_ptr<SSL_CTX, SSLCtxDeleter> ctx;
```

harryz2000: Sure. Will do

harryz2000: Done

src/TlsClient.cpp
```diff
@@ -0,0 +1,232 @@
...
+}
+
+TlsClient::~TlsClient()
+{
+	Shutdown();
+}
+
+bool TlsClient::Initialize(const char* hostname)
+{
+	ctx = std::unique_ptr<SSL_CTX, SSLCtxDeleter>(SSL_CTX_new(TLS_method()));
```
murillo128, line 35:
```suggestion
	ctx.reset(SSL_CTX_new(TLS_method()));
```

murillo128 (approved):
lgtm minor comments

PR #298, opened 2024-06-24
DIOS-5878 Add function for setting FrameDispatchCoordinator

harryz2000:


src/FrameDelayCalculator.cpp
```diff
@@ -1,5 +1,10 @@
 #include "FrameDelayCalculator.h"
 #include <iostream>
+#include <chrono>
+#include "log.h"
+
+using namespace std::chrono;
```
bcostdolby, line 6: Super minor by why do we need this? Most places in the code are using std::chrono:: still

src/MediaFrameListenerBridge.cpp
```diff
@@ -166,10 +166,22 @@ void MediaFrameListenerBridge::RemoveListener(RTPIncomingMediaStream::Listener*
 	});
 }
 
+void MediaFrameListenerBridge::SetFrameDispatchCoordinator(const std::shared_ptr<FrameDispatchCoordinator>& coordinator)
+{
+	timeService.Sync([=](auto now){
```
bcostdolby, line 171: Hard to see full context without all the repos in front of me. But are we sure it is safe to call Sync() here and wont lock up (and do we need Sync or is Async ok)?

bcostdolby (commented):
Not sure about some things. I dont have as much time to look into the things I commented on, are they important?

src/FrameDelayCalculator.cpp
harryz2000: This is used by `2000ms` at line 95. Yes, others don't use it. But it seems nicer to use `2000ms` rather than `std::chrono::2000ms`

src/MediaFrameListenerBridge.cpp
harryz2000: That is a good point. The caller must ensure it is called after the time service is started.

bcostdolby: Also needs to ensure the time service doesnt call Sync() on something that wants to run in the calling thread as well...

bcostdolby (approved):

include/FrameDispatchCoordinator.h
```diff
@@ -2,7 +2,10 @@
 #define FRAMEDISPATCHCOORDINATOR_H
 
 #include "FrameDelayCalculator.h"
-#include "MediaFrameListenerBridge.h"
+
+#include <mutex>
```
murillo128, line 6: We shouldn't use mutex on the media path. Try to use TimeServer::(A)Sync instead

murillo128: or at most atomic

harryz2000: Yeah. This is not ideal. However this has to be a sync call. I would improve this by only lock the two uint64_t values. The lock time must be very short. I think using atomic is probably better. 

There also seems to be ways to make 128bit access lock-free, but not looking into too deep. Such as discussed at. https://stackoverflow.com/questions/18177622/how-to-atomically-add-and-fetch-a-128-bit-number-in-c. Will check that.

murillo128: the locked time is king of irrelevant, the lock&free kills the performance.  why would we need an atomic for 128bit? if it is due to the chrono ms, let's use a uint64_t better

harryz2000: We need access the reference arrival time and reference timestamp at same time. Both are uint64_t values.

murillo128: yes, sue but we need only maxDelayMs to be atomic, rigth?

harryz2000: maxDelayMs can be atomic. Yes.

harryz2000: I created another PR to address the concern. https://github.com/medooze/media-server/pull/305
I made the maxDelayMs atomic. The challenge is to replace the mutex as FrameDelayCalculator::OnFrame is not thread safe. I will think about more about this.

PR #299, opened 2024-06-24
Fixed missing mutex unlock

bcostdolby:
…ase a mutex


harryz2000 (approved):
lgtm. But imho it would be better to use c++ lock/mutex with RAII.

bcostdolby:
Yeah going to create a different PR for that. I was going to change it all now but just want to get this in.

mildsunrise:
yeah, it would be nice to get rid of all the missing pthread_mutex usages in our codebase, at some point...

PR #300, opened 2024-06-24
DIOS-6081 Update VideoPipe to use C++ mutex and RAII

bcostdolby:


bcostdolby:
I already merged the quick fix using pthreads. This will hopefully prevent it from happening again but was a bigger change than I originally thought it would be.

src/VideoPipe.cpp
```diff
@@ -107,110 +93,83 @@ int VideoPipe::StartVideoCapture(uint32_t width, uint32_t height, uint32_t fps)
...
 			{
-				//wait
-				ret = pthread_cond_timedwait(&newPicCond,&newPicMutex,&wakeupTime);
-				if (ret == ETIMEDOUT)
+				if (!newPicCond.wait_until(newPicLock, wakeupTime.value(), haveWork))
 				{
-					// On timeout exit the loop
-					break;
+					// On timeout return null
+					return videoBuffer;
```
murillo128, line 159: wouldn't this change the behavior if downrating with a timeout? previously we would use last frame, now we would return null instead.

murillo128: @bcostdolby ?

bcostdolby: I will look further tomorrow but I think this is the current behaviour in master not a change (though we probably did change this behaviour sometime from the original when doing the pipe updates).

I.e. We currently do not support downrating by fractional amounts as we wanted to avoid duplicate timestamps. 

We mentioned this as a area to improve by creating new timestamps based on the rate and ingest timestamp sequence so if fractional will be identical and if not will interpolate. But thats a future task after the audio transcode fix.

PR #305, opened 2024-07-04
Remove lock for frame delay cacluation

harryz2000:


murillo128:
I think it would be easier to add a TimeService to the coordinator and call the OnFrame and SetMaxDelayMs on an Async call. The MediaFrameBridgeListener call is already thread safe.

harryz2000:
> I think it would be easier to add a TimeService to the coordinator and call the OnFrame and SetMaxDelayMs on an Async call. The MediaFrameBridgeListener call is already thread safe.

Yes. That would solve the thread lock issue. But to me, it seems not ideal as the calculated delay may not be used for current frame. For cases such as receive a batch of audio frames, they may still use a same delay, which might not be smooth.

Although, in practice, it might not be a issue. I am trying to see if it could be solved by another way. Got some progress, will need spend a bit more time to ensure it works.

harryz2000:
This should be a draft full solution for this issue. There is still a test to fix. But it should work. 

It calculates the delay in synchronously without lock. On the other hand it reduces latency asynchronously if all frames come early. Now the OnFrame is thread safe.

Yes. It has been a bit harder to implement this. But imho it suits better for our expectation.

harryz2000:
@murillo128 Now all unit tests have passed. I did some manual tests well. It works well to me, so I think it is ready for formal review. Personally, I would prefer this solution. It actually becomes cleaner now: One part is for delay calculation, which is lock-free sync operations. Another part for latency reduction, which uses async calls.

The test is using a TestTimeService which does sync calls for Async(), therefore most of the tests passed. The only test failed and has been fixed is the latency reduction may be delayed by one frame, as we are not reduce the delay immediately. However it shouldn't matter.

murillo128 (changes requested):

include/FrameDelayCalculator.h
```diff
@@ -26,16 +28,33 @@ class FrameDelayCalculator
 	std::chrono::milliseconds OnFrame(uint64_t streamIdentifier, std::chrono::milliseconds now, uint64_t ts, uint64_t clockRate);
 
 private:
+
+	union alignas (16) Reference
```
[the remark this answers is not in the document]
murillo128: @harryz2000 ptal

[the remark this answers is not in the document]
murillo128: @harryz2000 ping?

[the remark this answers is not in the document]
harryz2000: If we use  a class only for the fields, it would be like:
```
struct Reference
{
		__uint128_t value;
		struct
		{
			int64_t refTime;
			uint64_t refTimestamp;
		} content;
}
```

Then we still need to unpack/pack the value from/to the content. Using a uninon is convenient as it is automatically converted and no need extra memory.

I would create an up class to wrap the function, which might make things simpler.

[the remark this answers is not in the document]
harryz2000: Using union is convenient for pack/unpack the 128bit data. I would create a class to wrap the write/read calls.

src/FrameDelayCalculator.cpp
```diff
@@ -18,20 +18,41 @@ namespace
...
+static constexpr T ConvertTimestampClockRate(T ts, uint64_t originalRate, uint64_t targetRate)
 {
 	static_assert(sizeof(T) >= 8);
 	return originalRate == targetRate ? ts : (ts * T(targetRate) / T(originalRate));
 }
 
+/**
+ * Atomatically write a 128bit integer to an address
+ */
+void SyncWriteUint128(__uint128_t *dst, __uint128_t value)
```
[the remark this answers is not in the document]
harryz2000: sure

include/FrameDelayCalculator.h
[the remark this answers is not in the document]
harryz2000: Done

include/FrameDelayCalculator.h
```diff
@@ -27,15 +29,42 @@ class FrameDelayCalculator
 
 private:
 	
+	class Reference
+	{
+	public:
+		std::pair<int64_t, uint64_t> Get();
```
murillo128, line 35: could we use std::chrono::milliseconds on the getter/setters to remove the coversions on the main code?

test/unit/TestFrameDelayCalculator.cpp
```diff
@@ -56,26 +58,28 @@ void TestFrameDelayCalculator::TestDelayCalculator(const std::vector<std::tuple<
 	std::vector<std::pair<uint64_t, int64_t>> delays;
 
 	// Loop through all frames
+	int cnt = 0;
```
murillo128, line 61: is this variable used at all?

murillo128 (approved):
lgtm

test/unit/TestFrameDelayCalculator.cpp
murillo128, line 61:
```suggestion
```

include/FrameDelayCalculator.h
```diff
@@ -27,15 +29,42 @@ class FrameDelayCalculator
...
+		void Set(int64_t refTime, uint64_t refTimestamp);
```
murillo128, line 36:
```suggestion
		void Set(std::chrono::milliseconds refTme, uint64_t refTimestamp);
```

include/FrameDelayCalculator.h
murillo128:
```suggestion
		std::pair<std::chrono::milliseconds, uint64_t> Get();
```

src/FrameDelayCalculator.cpp
```diff
@@ -42,20 +43,20 @@ std::chrono::milliseconds FrameDelayCalculator::OnFrame(uint64_t streamIdentifie
...
+	
+	auto [refTime, refTimestamp] = reference.Get();
+	
+	if (refTime == 0 && refTimestamp == 0)
 	{
-		refTime = now;
-		refTimestamp = unifiedTs;
-		
-		initialized = true;			
+		reference.Set(now.count(), unifiedTs);
```
murillo128, line 52:
```suggestion
		reference.Set(now, unifiedTs);
```

src/FrameDelayCalculator.cpp
```diff
@@ -42,20 +43,20 @@ std::chrono::milliseconds FrameDelayCalculator::OnFrame(uint64_t streamIdentifie
...
 		return 0ms;
 	}
 			
 	// Note the lateMs could be negative when the frame arrives earlier than scheduled	
-	auto lateMs = GetFrameArrivalDelayMs(now, unifiedTs);
-
+	auto lateMs = GetFrameArrivalDelayMs(now, unifiedTs,
+				std::chrono::milliseconds(refTime), refTimestamp);
```
murillo128, line 58:
```suggestion
				refTime, refTimestamp);
```

src/FrameDelayCalculator.cpp
```diff
@@ -73,77 +74,111 @@ std::chrono::milliseconds FrameDelayCalculator::OnFrame(uint64_t streamIdentifie
 		Warning("-FrameDelayCalculator::OnFrame() | Timestamp clock jump detected [diff:%llu]\n",timestampDiff);
 		allEarlyStartTimeMs.reset();
 		frameArrivalInfo.erase(streamIdentifier);
-		refTime = 0ms;
+	
+		reference.Set(0, refTimestamp);
```
murillo128, line 78:
```suggestion
		reference.Set(0ms, refTimestamp);
```

src/FrameDelayCalculator.cpp
```diff
@@ -73,77 +74,111 @@ std::chrono::milliseconds FrameDelayCalculator::OnFrame(uint64_t streamIdentifie
...
 		return 0ms;
 	}
 	//If we were reseted because of a ts jump
-	else if (!refTime.count())
+	else if (!refTime)
 	{
 		//Store valid timestamp after reset
-		refTime = now;
+		refTime = now.count();
```
murillo128, line 85:
```suggestion
		refTime = now;
```

src/FrameDelayCalculator.cpp
```diff
@@ -73,77 +74,111 @@ std::chrono::milliseconds FrameDelayCalculator::OnFrame(uint64_t streamIdentifie
...
 		refTimestamp = unifiedTs;
+		
+		reference.Set(now.count(), unifiedTs);
```
murillo128, line 88:
```suggestion
		reference.Set(now, unifiedTs);
```

src/FrameDelayCalculator.cpp
```diff
@@ -73,77 +74,111 @@ std::chrono::milliseconds FrameDelayCalculator::OnFrame(uint64_t streamIdentifie
...
 	}
 
-    if (lateMs > updateRefsPacketLateThresholdMs)  // Packet late
+	bool early = false;
+	if (lateMs > updateRefsPacketLateThresholdMs)  // Packet late
 	{
-		refTime = now;
-		refTimestamp = unifiedTs;
+		reference.Set(now.count(), unifiedTs);
```
murillo128, line 94:
```suggestion
		reference.Set(now, unifiedTs);
```

src/FrameDelayCalculator.cpp
```diff
@@ -73,77 +74,111 @@ std::chrono::milliseconds FrameDelayCalculator::OnFrame(uint64_t streamIdentifie
...
+		
 		// Loop to see if all the streams have arrived earlier
-		allEarly = std::all_of(frameArrivalInfo.begin(), frameArrivalInfo.end(), 
+		bool allEarly = std::all_of(self->frameArrivalInfo.begin(), self->frameArrivalInfo.end(), 
 			[&](const auto& info) {
 				if (info.first == streamIdentifier) return true;
 							
 				auto [time, timestamp] = info.second;
-				auto frameLateMs = GetFrameArrivalDelayMs(time, timestamp);
+				auto frameLateMs = GetFrameArrivalDelayMs(time, timestamp, std::chrono::milliseconds(refTime), refTimestamp);
```
murillo128, line 124:
```suggestion
				auto frameLateMs = GetFrameArrivalDelayMs(time, timestamp, refTime, refTimestamp);
```

src/FrameDelayCalculator.cpp
```diff
@@ -73,77 +74,111 @@ std::chrono::milliseconds FrameDelayCalculator::OnFrame(uint64_t streamIdentifie
...
+std::pair<int64_t, uint64_t> FrameDelayCalculator::Reference::Get()
+{
+	static_assert(sizeof(ReferenceField) == 16);
+	static_assert(sizeof(ReferenceField::value) == 16);
+	static_assert(sizeof(ReferenceField::content) == 16);
+	
+	ReferenceField currentRef;
+	currentRef.value = __sync_fetch_and_add(&field, 0);
+	
+	return { currentRef.content.refTime, currentRef.content.refTimestamp };
```
murillo128, line 174:
```suggestion
	return { std::chrono::millisecondscurrentRef.content.refTime), currentRef.content.refTimestamp };
```

src/FrameDelayCalculator.cpp
```diff
@@ -73,77 +74,111 @@ std::chrono::milliseconds FrameDelayCalculator::OnFrame(uint64_t streamIdentifie
...
 	constexpr int64_t MsToTimestampFactor = UnifiedClockRate / 1000;
 	auto scheduledMs = (int64_t(unifiedTs) - int64_t(refTimestamp)) / MsToTimestampFactor + refTime.count();
 	auto actualTimeMs = now.count();
 
 	return actualTimeMs - scheduledMs;
-}
\ No newline at end of file
+}
+
+std::pair<int64_t, uint64_t> FrameDelayCalculator::Reference::Get()
```
murillo128, line 165:
```suggestion
std::pair<std::chrono::milliseconds, uint64_t> FrameDelayCalculator::Reference::Get()
```

src/FrameDelayCalculator.cpp
```diff
@@ -73,77 +74,111 @@ std::chrono::milliseconds FrameDelayCalculator::OnFrame(uint64_t streamIdentifie
...
+	static_assert(sizeof(ReferenceField::value) == 16);
+	static_assert(sizeof(ReferenceField::content) == 16);
+	
+	ReferenceField currentRef;
+	currentRef.value = __sync_fetch_and_add(&field, 0);
+	
+	return { currentRef.content.refTime, currentRef.content.refTimestamp };
+}
+
+void FrameDelayCalculator::Reference::Set(int64_t refTime, uint64_t refTimestamp)
```
murillo128, line 177:
```suggestion
void FrameDelayCalculator::Reference::Set(std::chrono::milliseconds refTime, uint64_t refTimestamp)
```

src/FrameDelayCalculator.cpp
```diff
@@ -73,77 +74,111 @@ std::chrono::milliseconds FrameDelayCalculator::OnFrame(uint64_t streamIdentifie
...
+{
+	ReferenceField newRef;
+	newRef.content = {refTime, refTimestamp};
```
murillo128, line 180:
```suggestion
	newRef.content = {refTime.count(), refTimestamp};
```

src/FrameDelayCalculator.cpp
```diff
@@ -73,77 +74,111 @@ std::chrono::milliseconds FrameDelayCalculator::OnFrame(uint64_t streamIdentifie
...
+	
+	ReferenceField currentRef;
+	currentRef.value = __sync_fetch_and_add(&field, 0);
+	
+	return { currentRef.content.refTime, currentRef.content.refTimestamp };
+}
+
+void FrameDelayCalculator::Reference::Set(int64_t refTime, uint64_t refTimestamp)
+{
+	ReferenceField newRef;
```
murillo128, line 179: Would this work?
```suggestion
	ReferenceField newRef = {{{refTime.count(), refTimestamp}}};
```

include/FrameDelayCalculator.h
harryz2000: Yes. We can

src/FrameDelayCalculator.cpp
harryz2000: Yes. We can do similar thing as that. But need the order of the fields changed. Done.

harryz2000:
@murillo128 Please have a look at this commit: https://github.com/medooze/media-server/pull/305/commits/4ec2a6ea509dabc73bd52ed9e7f82e6c5885bb19
I tweaked the code as currently the FrameArriveInfo would only be changed asynchronously. Also added "State" to clarify the case.

include/FrameDelayCalculator.h
```diff
@@ -1,20 +1,22 @@
...
+#include "TimeService.h"
+
 #include <unordered_map>
 #include <optional>
 #include <chrono>
 #include <queue>
 #include <algorithm>
 
-class FrameDelayCalculator
+class FrameDelayCalculator : public std::enable_shared_from_this<FrameDelayCalculator>
```
yezhan10, line 12: maybe irrelevant, what's the benefit of inheriting enabled_shared_from_this? i dont see shared_from_this or weak_from_this being used in framedelaycalculator.

harryz2000: We do use the weak_from_this() in the implement file. See line 91.
`timeService.Async([selfWeak = weak_from_this(), early, now, unifiedTs, refTime, refTimestamp,`

PR #307, opened 2024-07-11
Draft: DO NOT MERGE Hacks to enable debugging and dynamic config

bcostdolby:
This is not expected to be merged into master. Or to be reviewed as such. This branch is adding a bunch of things to help us debug issues more quickly if they appear and try modifying things dynamically if they arise for the demo.

Please let me know if there is anything in here that might cause instability or crashes. I also dont think we will use this when we get the the point of scaling up the demo. 

I have added:
* Logging messages with the viewer ID so we know what logs are for what viewers by filtering on that
* Add configuration of some items that might permit BWE or other things to be less aggressive (windows sizes, increase rates etc)
* There are also lots of questions in the code marked as @todo, largely why we choose max/min in certain cases.

### DIFF
--- a/include/DTLSICETransport.h
+++ b/include/DTLSICETransport.h
@@ -67,7 +67,7 @@ public:
 	};
 
 public:
-	DTLSICETransport(Sender *sender,TimeService& timeService, ObjectPool<Packet>& packetPool);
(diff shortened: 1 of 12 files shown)


bcostdolby:
Splitting this into separate PRs for things that are relevant

PR #308, opened 2024-07-16
add av1 decoder support

yezhan10:
add dav1d av1 decoder

### DIFF
--- a/src/VideoCodecFactory.cpp
+++ b/src/VideoCodecFactory.cpp
@@ -10,6 +10,7 @@
 #include "jpeg/JPEGEncoder.h"
 #include "webp/WEBPEncoder.h"
 #include "av1/AV1Encoder.h"
+#include "av1/AV1Decoder.h"
 
 VideoDecoder* VideoCodecFactory::CreateDecoder(VideoCodec::Type codec)
 {
@@ -26,6 +27,8 @@ VideoDecoder* VideoCodecFactory::CreateDecoder(VideoCodec::Type codec)
 			return new VP8Decoder();
 		case VideoCodec::VP9:
 			return new VP9Decoder();
+		case VideoCodec::AV1:
+			return new AV1Decoder();
 		default:
 			Error("Video decoder not found [%d]\n",codec);
 	}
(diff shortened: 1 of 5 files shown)


src/av1/AV1decoder.cpp
```diff
@@ -0,0 +1,143 @@
...
+		case AVCOL_RANGE_JPEG: 
+			//2^n-1   "JPEG" YUV ranges
+			videoBuffer->SetColorRange(VideoBuffer::ColorRange::Full);
+			break;
+		default:
+			//Unknown
+			videoBuffer->SetColorRange(VideoBuffer::ColorRange::Unknown);
+	}
+	
+	// TODO: color space?
```
phil-moss-dolby, line 130: Should be the same color spaces as h264, in addition AV1 can handle HDR10 and HDR metadata. I just don't know if we're adding those cause I don't see them in h265decoder

src/VideoCodecFactory.cpp
```diff
@@ -9,6 +9,7 @@
 #include "vp9/VP9Encoder.h"
 #include "jpeg/JPEGEncoder.h"
 #include "webp/WEBPEncoder.h"
+#include "av1/AV1decoder.h"
```
murillo128, line 12: Pleas rename

```suggestion
#include "av1/AV1Decoder.h"
```

murillo128 (approved):
lgtm, but please rename the AV1Decoder files to match the class name case

src/av1/AV1decoder.cpp
yezhan10: new commit has color space set as the same as h264

murillo128 (approved):
replace space idents on test file by tabs, otherwise lgtm. Good job!

.github/workflows/release.yml
```diff
@@ -20,6 +20,9 @@ jobs:
       run: |
         sudo apt-get update
         sudo apt-get install --fix-missing -y libavcodec-dev libx264-dev libx265-dev libavutil-dev libvpx-dev libswscale-dev libswresample-dev libavformat-dev libssl-dev libavfilter-dev
+        sudo wget https://ftp.debian.org/debian/pool/main/a/aom/libaom3_3.9.1-1_amd64.deb
```
murillo128, line 23: I have merged the AV1 encoder support, so we can remove this from here

PR #309, opened 2024-07-16
DIOS-6296 Allow video pipe to pass through old TS frames if encoder resets

bcostdolby:


murillo128 (approved):
lgtm, we should add some unit tests for this.

murillo128 (approved):
lgtm

PR #310, opened 2024-07-16
Detect timestamp jumps

murillo128:
This PR detects jumps in the clock reset and timestamp jumps on the FrameDelayCalculator and avoid increasing the delay in that case.

### DIFF
--- a/src/FrameDelayCalculator.cpp
+++ b/src/FrameDelayCalculator.cpp
@@ -8,6 +8,7 @@ using namespace std::chrono;
 namespace
 {
 
+
 /**
  * Convert timestamp from one clock rate to another
  * 
@@ -25,6 +26,7 @@ static constexpr T convertTimestampClockRate(T ts, uint64_t originalRate, uint64
 }
 
 static constexpr uint64_t UnifiedClockRate = 90 * 1000;
+static constexpr uint64_t MaxClockDesync = 100 * UnifiedClockRate; //100s
 }
 
 FrameDelayCalculator::FrameDelayCalculator(int aUpdateRefsPacketLateThresholdMs, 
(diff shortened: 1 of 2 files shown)


murillo128:
@harryz2000 @bcostdolby  please review it even if it is merged

src/FrameDelayCalculator.cpp
```diff
@@ -48,23 +50,46 @@ std::chrono::milliseconds FrameDelayCalculator::OnFrame(uint64_t streamIdentifie
...
+	{
+		//Reset calculator, we keep the refTimestamp so it is not set to the stream that has the jump
+		Warning("-FrameDelayCalculator::OnFrame() | Timestamp clock jump detected [diff:%llu]\n",timestampDiff);
+		allEarlyStartTimeMs.reset();
+		frameArrivalInfo.erase(streamIdentifier);
+		refTime = 0ms;
+		return 0ms;
+	}
+	//If we were reseted because of a ts jump
+	else if (!refTime.count())
```
bcostdolby, line 80: Not sure why this is an "else if "and not just an "if" I assume it has something to do with the refTimestamp only being reset

src/FrameDelayCalculator.cpp
```diff
@@ -48,23 +50,46 @@ std::chrono::milliseconds FrameDelayCalculator::OnFrame(uint64_t streamIdentifie
...
 	bool allEarly = false;
-	if (lateMs > updateRefsPacketLateThresholdMs)  // Packet late
+	//If we detect a clock difference between the timestamps
+	if (refTimestamp && timestampDiff > MaxClockDesync)
+	{
+		//Reset calculator, we keep the refTimestamp so it is not set to the stream that has the jump
+		Warning("-FrameDelayCalculator::OnFrame() | Timestamp clock jump detected [diff:%llu]\n",timestampDiff);
+		allEarlyStartTimeMs.reset();
+		frameArrivalInfo.erase(streamIdentifier);
+		refTime = 0ms;
```
bcostdolby, line 76: I didnt fully understand why refTimestamp is not reset inside this if statement along with refTime 

If we have multiple streams and then 1 of them jumps back (or forward), we are resetting the state but should we reset it for all of them. I.e. The expectation is that they all are roughly in sync as that is what it expects. 

Maybe more comments on why? The comment makes it clear there is a reason I dont know

src/FrameDelayCalculator.cpp
murillo128: as the `if` is doing a `return 0`  the `else if` could be an  `if``, there should be no difference though

src/FrameDelayCalculator.cpp
murillo128: we don't want to reset the `refTimestamp` as if not the next frame will set which clock is valid. Keeping `refTimestamp` value means that we only set the diff to 0 for the stream that has the clock jump

PR #312, opened 2024-08-12
DIOS-6494 Audio transcoding refactor step 1: update audioDecoder

yezhan10:
1. Update AudioDecoder::Decode() implementation that takes a AudioFrame and returns PCM data
2. Update this implementation to all audio decoders that we support
3. Added unit test for this (only aacDecoder for now)

### DIFF
--- a/src/g711/pcmacodec.cpp
+++ b/src/g711/pcmacodec.cpp
@@ -38,10 +38,13 @@ int PCMAEncoder::Encode (SWORD *in,int inLen,BYTE* out,int outLen)
 	return inLen;
 }
 
-int PCMADecoder::Decode (const BYTE *in,int inLen,SWORD* out,int outLen)
+int PCMADecoder::Decode(const AudioFrame::const_shared& audioFrame, SWORD* out,int outLen)
 {
 	//Comprobamos las longitudes
-	if (outLen<inLen)
+	// inLen: compressed data length in bytes/samples one byte per sample for g711a, outLen: pcm data size in samples
+	int inLen = audioFrame->GetLength();
+	const uint8_t* in = audioFrame->GetData();
+	if (outLen<inLen || !in)
 		return 0;
 
 	//Decodificamos
--- a/src/g711/pcmucodec.cpp
+++ b/src/g711/pcmucodec.cpp
@@ -37,10 +37,13 @@ int PCMUEncoder::Encode (SWORD *in,int inLen,BYTE* out,int outLen)
 	return inLen;
 }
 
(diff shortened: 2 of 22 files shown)


src/g711/pcmacodec.cpp
```diff
@@ -38,10 +38,13 @@ int PCMAEncoder::Encode (SWORD *in,int inLen,BYTE* out,int outLen)
...
 
-int PCMADecoder::Decode (const BYTE *in,int inLen,SWORD* out,int outLen)
+int PCMADecoder::Decode(const AudioFrame::const_shared& audioFrame, SWORD* out,int outLen)
 {
 	//Comprobamos las longitudes
-	if (outLen<inLen)
+	// inLen: compressed data length in bytes, outLen: data size in samples
+	int inLen = audioFrame->GetLength();
+	const uint8_t* in = audioFrame->GetData();
+	if (outLen<inLen || !in)
```
bcostdolby, line 47: Thanks for adding the comment. Will approve for now, but this check still makes no sense. We need to check that the decoded length wont exceed the outlen but we can deal with that in another ticket.

bcostdolby (approved):
lgtm

src/g711/pcmacodec.cpp
mildsunrise: this check makes sense to me. input samples are 8-bit, output samples are 16-bit, right?

mildsunrise (approved):
yayyyy, it's finally happening :blue_heart: thank you. lgtm

src/g711/pcmacodec.cpp
bcostdolby: I missed that this was a g711 a/u specific check that doesn't exist in other codecs. Yes that makes sense. We should change the comment. I.e. In this case a byte == a sample

// inLen: compressed data length in bytes/samples one byte per sample for g711a, outLen: pcm data size in samples

src/g711/pcmucodec.cpp
```diff
@@ -37,10 +37,13 @@ int PCMUEncoder::Encode (SWORD *in,int inLen,BYTE* out,int outLen)
 	return inLen;
 }
 
-int PCMUDecoder::Decode (const BYTE *in,int inLen,SWORD* out,int outLen)
+int PCMUDecoder::Decode(const AudioFrame::const_shared& audioFrame, SWORD* out,int outLen)
 {
-	//Comprobamos las longitudes
-	if (outLen<inLen)
+	// Comprobamos las longitudes
+	// inLen: compressed data length in bytes, outLen: data size in samples
```
bcostdolby, line 43: Also update this comment

murillo128 (approved):
lgtm, great job!

PR #313, opened 2024-08-13
DIOS-6105 Return rectified timestamp in timestamp checker

harryz2000:
The purpose of this PR is to provide facility to allow the caller to get a continuous and incremental timestamp even when invalid timestamp was found or timestamp base had changed. 

The change include:
1. Change timestamp checker to return rectified timestamp as per receiving time in case of invalid timestamp is detected. 
2. Introduce an offset in timestamp checker so the offset can be changed in case of a reference reset after continuous invalid timestamp were detected.

As the timestamp was calculated as per receiving time, it should always not far away from the expected ideal timestamp. And it is always incremental. Note the default timestamp tolerance is still available, which means a timestamp wouldn't be regarded invalid is there is small variation. 

If timestamp base changed, the initial few timestamps would be regarded as invalid (corrected timestamp would still be returned). Then the timechecker reference would be reset and the offset would be changed as per the new timestamps. Therefore, the following timestamp would be regarded as valid, although the offset would be used to calculate the correct timestamp.


include/TimestampChecker.h
```diff
@@ -59,30 +62,45 @@ class TimestampChecker
...
-			numContinousInvalidFrames = 0;
 			
-			return CheckResult::Reset;
-		}
+			rectifiedTs = (int64_t(recvTimeMs) - int64_t(lastRecvTime))* clock / 1000 + lastTimeStamp + offset;
+
+			// If there are continous invalid frames, it could be due to reference wrong. Reset to current.
+			if (numContinousInvalidFrames > maxContinousInvalidFrames)
+			{
+				lastTimeStamp = timestamp;
```
bcostdolby, line 88: I am wondering if we need a logger in here or if we can log sufficient detail to debug issues in the caller. Either way, whenever we reset the sequence we need to make sure we log a warning (ok with it in the caller).

include/TimestampChecker.h
```diff
@@ -59,30 +62,45 @@ class TimestampChecker
...
-		// If there are continous invalid frames, it could be due to reference wrong. Reset to current.
-		if (numContinousInvalidFrames > maxContinousInvalidFrames)
-		{
-			lastTimeStamp = timestamp;
-			lastRecvTime = recvTimeMs;
-			numContinousInvalidFrames = 0;
 			
-			return CheckResult::Reset;
-		}
+			rectifiedTs = (int64_t(recvTimeMs) - int64_t(lastRecvTime))* clock / 1000 + lastTimeStamp + offset;
```
bcostdolby, line 83: I do wonder if this is the best way to do it. Its probably ok, but if we have 25 fps for example we would expect the ts to increment at 40msec each frame. This would be very consistent generally and using real-world times here means we will lose the consistency. 

This might be a good thing to be able to see what is going on, but not certain it will be 100% suitabe for all downstream timestamp handling code (probably ok).

bcostdolby (commented):
We should hold off on merging this until after the 2.19.0 release I think.

bcostdolby:
Please update the PR title to use the conventional commits format

include/TimestampChecker.h
harryz2000: Currently, the caller would be responsible for the logging. I thought of adding logs here. However, this class is a very generic class and doesn't contains some identification information. The caller would have more information to log.

include/TimestampChecker.h
harryz2000: One issue is the frames may actually have gaps. For example, in the real pcap file. The timestamp jump happened while the receiving time is about 9 seconds after the previous frame. In that case, using a single duration might not be correction. We may have to use receiving time to guess. 

Yes. This is not the perfect way. But that is the most reasonable solution I found so far.

murillo128: calculating the received time offset for adjusting the ts is the only way we can have continuous timestamp when the sources are switched. 
The only thing I am worried is how to take into account congestion delay, as the `recvTimeMs` may lagging behing the actual frame time, which will cause the `rectifiedTs` to be smaller than it should be. Probably something we can leave with though.

murillo128 (commented):
There is one thing I would like to change though
```             if (valid)
		{
			return { CheckResult::Valid, rectifiedTs };
		}
		else
		{
			if (numContinousInvalidFrames > maxContinousInvalidFrames)
			{
				return { CheckResult::Reset, rectifiedTs};
			}
			return { CheckResult::Invalid, rectifiedTs };
		}

File: include/TimestampChecker.h (hunk 2 above)
Comment (harryz2000) [reply include/TimestampChecker.h:83]: Yes. The calculation may result in the timestamp to be different with the ideal one. But fortunately these calculation only lasts a few frames. Once we reset the offset, the rectifiedTs would be basing one the original timestamp.

**Comment by harryz2000** [comment]
Resolved all the comments.

File: include/TimestampChecker.h [hunk 3]
```diff
@@ -59,30 +77,50 @@ class TimestampChecker
...
+		
+		return { result, uint64_t(rectifiedTs) };
+	}
+	
+	inline int64_t GetTimestampOffset() const
+	{
+		return offset;
 	}
 	
 	inline static constexpr uint64_t GetDiff(uint64_t lhs, uint64_t rhs)
```
murillo128, line 126: When reading the code It is a bit confusing that the dif is the absolute diff:
```suggestion
	inline static constexpr uint64_t GetAbsDiff(uint64_t lhs, uint64_t rhs)
```

include/TimestampChecker.h
```diff
@@ -59,30 +77,50 @@ class TimestampChecker
 		// The difference between expected duration as per timestamp and actual duration since last frame
 		auto diff = sameOrder ? GetDiff(durationOnTs, durationOnRecv) : (durationOnTs + durationOnRecv);
 		
+		auto result = CheckResult::Valid;
+		int64_t rectifiedTs = 0;
```
murillo128, line 81: the `rectifiedTs` should be the unmodified timestamp by default

include/TimestampChecker.h
```diff
@@ -59,30 +77,50 @@ class TimestampChecker
...
 
-		return valid ? CheckResult::Valid : CheckResult::Invalid;
+			// If there are continous invalid frames, it could be due to reference wrong. Reset to current.
+			if (numContinousInvalidFrames > maxContinousInvalidFrames)
+			{
+				lastTimeStamp = timestamp;
+				lastRecvTime = recvTimeMs;
+				numContinousInvalidFrames = 0;
+				
+				offset = rectifiedTs - timestamp;
```
murillo128, line 108: what do we use the offset for?

harryz2000: The offset is set when a reset happens. So after reset we would use received timestamp + offset, instead of using received time to calculate. The offset is to ensure the timestamp incremental

murillo128: if we are not using offset anywhere else, ie we are using it only for `lastTimestamp + offset`, which is the last `rectifiedTimestamp`, wouldn't it be better to just store the `lastRectifedTimestamp` instead?

harryz2000: We use the offset for following "valid" timestamps after reset, at line 91.

In this way, if the following timestamp is considered as valid after resetting, the original `timestamp + offset` would be used as rectified timestamp. This would be more accurate than the timestamp generated from receiving time.

include/TimestampChecker.h
harryz2000: Done.

murillo128 (approved):
lgtm

PR #314, opened 2024-08-13
No need to checkout mp4v2 to a git controlled folder

harryz2000:
Checking out to a git controlled folder always make it regarded as a uncommitted change.
Some dependency improvement


bcostdolby (approved):
lgtm

PR #316, opened 2024-08-13
DIOS-6105 Add timestamp checking

harryz2000:
Use rectified timestamp in MediaFrameListenerBridge (for SRT and RTMP) and RTPIncomingMediaStreamDepacketizer (for WebRTC). 

Note a flag was added to MediaFrameListenerBridge to disable the timestamp check by default as the class is also used by transcoder. We would expect the timestamp from transcoder matches the input. So we don't need to check the timestamp again there.

### DIFF
--- a/include/MediaFrameListenerBridge.h
+++ b/include/MediaFrameListenerBridge.h
@@ -5,6 +5,7 @@
 #include "media.h"
 #include "rtp.h"
 #include "TimeService.h"
+#include "TimestampChecker.h"
 #include "FrameDispatchCoordinator.h"
 
 #include <queue>
@@ -37,7 +38,7 @@ public:
 	};	
 	
 public:
-	MediaFrameListenerBridge(TimeService& timeService, DWORD ssrc, bool smooth = false);
+	MediaFrameListenerBridge(TimeService& timeService, DWORD ssrc, bool smooth = false, bool checkTimestamp = false);
 	virtual ~MediaFrameListenerBridge();
 
 	void SetFrameDispatchCoordinator(const std::shared_ptr<FrameDispatchCoordinator>& coordinator);
@@ -130,6 +131,7 @@ public:
 
 	uint32_t targetBitrateHint = 0;
 	
+	std::unique_ptr<TimestampChecker> tsChecker;
 	std::shared_ptr<FrameDispatchCoordinator> coordinator;
 };
 
--- a/include/MediaFrameListenerBridge.h
+++ b/include/MediaFrameListenerBridge.h
@@ -132,6 +132,7 @@ public:
 	uint32_t targetBitrateHint = 0;
 	
 	std::unique_ptr<TimestampChecker> tsChecker;
+	std::unique_ptr<TimestampChecker> ptsChecker;
 	std::shared_ptr<FrameDispatchCoordinator> coordinator;
 };
 
--- a/src/MediaFrameListenerBridge.cpp
+++ b/src/MediaFrameListenerBridge.cpp
@@ -184,7 +184,8 @@ void MediaFrameListenerBridge::onMediaFrame(DWORD ignored, const MediaFrame& fra
 			auto [status, rts] = tsChecker->Check(frame->GetTime(), frame->GetTimeStamp(), frame->GetClockRate());
 			if (status != TimestampChecker::CheckResult::Valid)
 			{
-				Error("Invalid timestamp. status: %d, info: %s, corrected: %llu\n", int(status), frame->TimeInfoToString().c_str(), rts);
+				Error("Invalid timestamp. status: %d, info: %s, corrected: %llu offset: %lld \n", 
+					TimestampChecker::CheckResultToString(status), frame->TimeInfoToString().c_str(), rts, tsChecker->GetTimestampOffset());
 			}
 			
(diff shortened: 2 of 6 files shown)


src/MediaFrameListenerBridge.cpp
```diff
@@ -177,6 +178,29 @@ void MediaFrameListenerBridge::onMediaFrame(DWORD ignored, const MediaFrame& fra
 {
 	timeService.Async([=, frame = std::shared_ptr<MediaFrame>(frame.Clone())] (auto now){
 		
+		if (tsChecker)
+		{
+			// Check timestamp
+			auto [status, rts] = tsChecker->Check(frame->GetTime(), frame->GetTimeStamp(), frame->GetClockRate());
+			if (status != TimestampChecker::CheckResult::Valid)
+			{
+				Error("Invalid timestamp. status: %d, info: %s, corrected: %llu\n", int(status), frame->TimeInfoToString().c_str(), rts);
```
bcostdolby, line 187: Minor: Can we log the status as a enum string to make the logs easier to read?

include/MediaFrameListenerBridge.h
```diff
@@ -37,7 +38,7 @@ class MediaFrameListenerBridge :
 	};	
 	
 public:
-	MediaFrameListenerBridge(TimeService& timeService, DWORD ssrc, bool smooth = false);
+	MediaFrameListenerBridge(TimeService& timeService, DWORD ssrc, bool smooth = false, bool checkTimestamp = false);
```
bcostdolby, line 41: I assume this is to prevent it being used for things like transcode layers? Usage code isnt in this PR

bcostdolby (commented):
Should wait until after 2.19.0 to merge IMO

bcostdolby:
Please update the PR title to use the conventional commits format

src/MediaFrameListenerBridge.cpp
harryz2000: Will do.

include/MediaFrameListenerBridge.h
harryz2000: Yes. It is to make it disabled by default

harryz2000:
> Please update the PR title to use the conventional commits format

Have we got one agreed format? :)

murillo128 (changes requested):

src/MediaFrameListenerBridge.cpp
```diff
@@ -177,6 +178,30 @@ void MediaFrameListenerBridge::onMediaFrame(DWORD ignored, const MediaFrame& fra
 {
 	timeService.Async([=, frame = std::shared_ptr<MediaFrame>(frame.Clone())] (auto now){
 		
+		if (tsChecker)
+		{
+			// Check timestamp
+			auto [status, rts] = tsChecker->Check(frame->GetTime(), frame->GetTimeStamp(), frame->GetClockRate());
+			if (status != TimestampChecker::CheckResult::Valid)
+			{
+				Error("Invalid timestamp. status: %d, info: %s, corrected: %llu offset: %lld \n",
```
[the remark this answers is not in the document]
harryz2000: The issue here is we don't enable warning level logging. If we change to warning, we would lose the information that is very important for debugging.

src/MediaFrameListenerBridge.cpp
```diff
@@ -177,6 +178,30 @@ void MediaFrameListenerBridge::onMediaFrame(DWORD ignored, const MediaFrame& fra
...
+			if (status != TimestampChecker::CheckResult::Valid)
+			{
+				Error("Invalid timestamp. status: %d, info: %s, corrected: %llu offset: %lld \n", 
+					TimestampChecker::CheckResultToString(status), frame->TimeInfoToString().c_str(), rts, tsChecker->GetTimestampOffset());
+			}
+			
+			if (frame->GetType() == MediaFrame::Video)
+			{
+				VideoFrame* vframe = static_cast<VideoFrame*>(frame.get());
+				auto pts = rts + vframe->GetPresentationTimestamp() - vframe->GetTimestamp();
```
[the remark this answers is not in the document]
harryz2000: The rationale is the pts is not far from ts and pts should be always >0. But yes this calculation has issues. I will add another checker for it.

src/rtp/RTPIncomingMediaStreamDepacketizer.cpp
```diff
@@ -40,6 +40,17 @@ void RTPIncomingMediaStreamDepacketizer::onRTP(const RTPIncomingMediaStream* gro
 	 //If we have a new frame
 	 if (frame)
 	 {
+		// Check timestamp
+		auto [status, rts] = tsChecker.Check(frame->GetTime(), frame->GetTimeStamp(), frame->GetClockRate());
```
murillo128, line 44: not sure if i want to put this code here, as it only would change the timestamps for the media frames, but not the timestamps for the rtp packets

harryz2000: That is right. We don't see a lot of timestamp issues for WebRTC except the rare random invalid timestamps. The check is critical for recordings, A/V sync, etc. But might not be that critical for WebRTC.

But we could change RTP packet timestamp as well.

PR #317, opened 2024-08-16
DIOS-5498 Change RTMPClientConnection to use EventLoop

harryz2000:


mildsunrise (commented):
I still need to take a deeper look but lgtm so far. always happy to move more of our code using the legacy `createPriorityThread` API into `EventLoop`

src/rtmp/rtmpclientconnection.cpp
```diff
@@ -362,10 +382,11 @@ void RTMPClientConnection::SignalWriteNeeded()
 	//Unlock
 	pthread_mutex_unlock(&mutex);
 
-	//Check thred
-	if (!isZeroThread(thread))
+	auto threadHandle = GetThread().native_handle();
```
murillo128, line 385: can't we just call `EventLoop::Signal()` instead?

murillo128 (commented):
lgtm, wondering if we can reuse more of the `EventLoop` functionality here though

src/rtmp/rtmpclientconnection.cpp
harryz2000: yeah. That works too.

murillo128: please change it then

murillo128 (changes requested):

src/rtmp/rtmpclientconnection.cpp
harryz2000: Yep. Haven't been able to come to this PR yet. Trying to resolve the rtmps first. This PR would be low priority as it would be not necessary but a good to have change as the timeservice here may not be used.

harryz2000: Done

harryz2000:
> lgtm, wondering if we can reuse more of the `EventLoop` functionality here though

I tried to reuse some members of EventLoop. But I feel it makes things a bit messy. I would prefer to keep the current change. 

Maybe on another PR, we could optimise the EventLoop a bit to make it easy to be extended. Such as if we could reuse the Run() function. If we could split the network functionality from a Basic EventLoop, i.e. Created a class such as NetEventLoop.  Those idea would need a further review of the code and a larger change.

murillo128 (approved):
lgtm

PR #319, opened 2024-08-21
DIOS-4033 Use fec packets for network probing

murillo128:
A version of https://github.com/medooze/media-server/pull/245 that doesn't change probing behaviour when flex-fec is not in use.


harryz2000:
It seems there is a issue causing the diff too large. May need to rebase?

murillo128:
rebased now

PR #321, opened 2024-08-23
Remove mcu files and move mixer components to its own dir

murillo128:
This PR removes old MCU service and moves the mixer components which may be reusable in the future to its own directory.


harryz2000 (approved):
lgtm

PR #328, opened 2024-08-28
Use exceptions for bit stream operations

murillo128:
This PR adds exceptions to the bit reader/writters and modifies the code that uses for catching them. This PR is done on top of #326


mildsunrise (approved):
tentative lgtm, I still have to review further (possibly once that is merged).
in the future we could propagate errors further from descriptor parsing rather than returning booleans.

include/bitstream/BitReader.h
```diff
@@ -339,16 +348,13 @@ class BitReader
 		}
 			
 	}
+
 	inline DWORD GetCached(DWORD n)
 	{
+		assert(n <= cached);
+
 		//Check length
 		if (!n) return 0;
```
bcostdolby, line 357: Should this be treated as an error as well (probably not but I wanted to raise the question)? Its a bit weird asking for 0 data unless a variable amount of bits needed, but then the 32 bit limit might be an issue as well unless the size is decoded from a few bits only.

include/bitstream/BitWritter.h
```diff
@@ -107,11 +105,15 @@ class BitWritter {
 	inline DWORD Put(BYTE n,DWORD v)
 	{
 		
-		if (!n) 
+		//Nothing to do
+		if (!n) return v;
+
+		//Check we have enougth space for writting n bits
+		if (n>Left())
+			throw std::range_error("not enougth buffer");
```
bcostdolby, line 113: typo for "enough"

include/bitstream/BitWritter.h
```diff
@@ -107,11 +105,15 @@ class BitWritter {
...
+
+		if (n+cached>32)
```
bcostdolby, line 115: I think bug in caller can overflow this to pass as i think it is all in BYTE so Put(254,v) and cached == 3 would pass.

src/aac/aacconfig.h
```diff
@@ -97,26 +97,34 @@ class AACSpecificConfig
...
+			writter.Put(1, frameLength); 
+			writter.Put(1, coreCoder); 
+			writter.Put(1, extension); 
+
+			//Flush
+			return writter.Flush();
+		}
+		catch (std::exception& e) 
+		{
+			return 0;
```
bcostdolby, line 126: Should we log these exceptions and below? It doesnt look like the callers always log errors? For writes at least it should be rare and not spammy and actually if we are seeing lots of corrupt input and it is spammy then we probably want to know about it.

Someday we will probably allow the exceptions to propagate out and give better info of the cause to the caller, but for now it may be useful to know if it was a logic or runtime error at least.

bcostdolby (commented):
I looked at a few caller use-cases and most seem to be only modifying temporaries which is good, but was unable to check all code paths.

include/bitstream/BitReader.h
```diff
@@ -70,16 +72,16 @@ class BitReader
 		cached = 0;
 		cache = 0;
 		bufferPos = 0;
-		//No error
-		error = false;
 	}
+
 	inline DWORD Get(DWORD n)
 	{
+		assert(n <= 32);
```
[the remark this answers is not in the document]
murillo128: We don't define NDEBUG, so assertions should be enabled on default build.

I don't have a strong oppinion on assert vs exception here, happy to use any of them

include/bitstream/BitReader.h
```diff
@@ -282,10 +296,7 @@ class BitReader
 			//Increase pointer
 			bufferPos++;
 		} else {
-			//We can't use exceptions so set error flag
-			error = true;
-			//Exit
-			return 0;
+			throw std::out_of_range("no more bytes to read from");
```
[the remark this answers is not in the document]
murillo128: what exception do  you think that would be better? I don't have a strong opinion, so happy to change to any other that is more relevant.

include/bitstream/BitReader.h
```diff
@@ -363,7 +369,6 @@ class BitReader
 	DWORD bufferPos;
 	DWORD cache;
 	BYTE  cached;
```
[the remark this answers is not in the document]
murillo128: I would prefer to do it in a following PR, this one is big enough. I would also like to change the buffer/bufferPos by a propper BufferReader, and maybe even allow to do RBSP escaping directly without doing a memory copy.

include/bitstream/BitReader.h
murillo128: asking for n=0 is just a noop which is harmless although maybe it is hiding a bug on the caller layer. Not sure what is the best solution.

include/bitstream/BitWritter.h
murillo128:
```suggestion
			throw std::range_error("not enough buffer");
```

src/aac/aacconfig.h
murillo128: Should not happen ever, but doesn't harm to add a warning.
```suggestion
			return Warning("-AACSpecificConfig:Serialize() | Could not serialize aac config\n");
```

include/bitstream/BitReader.h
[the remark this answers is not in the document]
bcostdolby: None of the std:: ones are perfect but derived from runtime_error is preferrable to logic_error, so maybe std::range_error?

Seems like it was merged anyway so doesnt really matter

PR #331, opened 2024-08-30
feat: add support for unegotiated incoming groups

murillo128:
This PR adds support for detecting unsignaled sources that provides relevant `mid` and `rid` values.

The `mid` is used for selecting the track to associate the id, and the `rid` for the source id. RTX ssrc is associated by `rid` or `rrid` afterwards.

Also I had to change some raw pointers by shared pointers.

### DIFF
--- a/include/DTLSICETransport.h
+++ b/include/DTLSICETransport.h
@@ -59,6 +59,7 @@ public:
 		virtual void onICETimeout() = 0;
 		virtual void onDTLSStateChanged(const DTLSState) = 0;
 		virtual void onRemoteICECandidateActivated(const std::string& ip, uint16_t port, uint32_t priority) = 0;
+		virtual void onUnsignaledIncomingSourceGroup(const RTPIncomingSourceGroup::shared& group) = 0;
 		virtual ~Listener() = default;
 	};
 	class Sender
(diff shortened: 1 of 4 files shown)


include/DTLSICETransport.h
```diff
@@ -179,10 +180,10 @@ class DTLSICETransport :
 	WORD		feedbackCycles			= 0;
 
 	//TODO: change by shared pointers
```
eck1001, line 182: can this todo be removed now?

include/rtp/RTPOutgoingSourceGroup.h
```diff
@@ -63,6 +63,7 @@ struct RTPOutgoingSourceGroup
 	}
 	
 public:	
+	std::string rid;
```
eck1001, line 66: question: rid is "repairedStreamId" correct? What caused an id to need repairing?

include/rtp/RTPOutgoingSourceGroup.h
```diff
@@ -63,6 +63,7 @@ struct RTPOutgoingSourceGroup
...
 	std::string mid;
```
eck1001, line 67: question: `mid` is mediaStreamId?

include/rtp/RTPOutgoingSourceGroup.h
murillo128: that's the name of the rtp header extension (https://www.rfc-editor.org/rfc/rfc8852.html#name-rtcp-repairedrtpstreamid-sd) for RTX.  rid is the rtp-stream-id and the rrid is the repaired-rtp-stream-id.

When receiving a RTX packate with an rrid value, you can find the media stream by checking  rtx rrid = media rid.

include/rtp/RTPOutgoingSourceGroup.h
murillo128: yes

include/DTLSICETransport.h
murillo128: yes!

PR #332, opened 2024-09-02
fix: [DIOS-6865] Add missing includes from (#328)

bcostdolby:
Note: Dont think I should go an rename BitWritter to BitWriter in this, just want to get things compiling again


harryz2000 (approved):
lgtm

PR #335, opened 2024-09-03
refactor: Modify audio codec API to use AudioBuffer

yezhan10:
What's included in this PR:
1. All audio encoders consume AudioBuffer, return AudioFrame pointer. 
2. All audio decoders consume AudioFrame, return AudioBuffer shared pointer obj. Since it's possible that a audio frame has multiple decoded PCM data. So split the original Decode() function to Decode() and GetDecodedAudioFrame(). 
3. Added unit tests for AudioPipe, OpusEncoder. 

Timestamp improvement will be included in next PR.


CMakeLists.txt
```diff
@@ -56,14 +56,14 @@ add_library(MediaServerLib
 	${SOURCES_WEBP}
 	${SOURCES_AV1}
     ${SOURCES_AAC}
-    ${SOURCES_MP3}
-    ${SOURCES_G711}
-    ${SOURCES_G722}
-    ${SOURCES_GSM}
+    # ${SOURCES_MP3}
```
murillo128, line 59: please revert

CMakeLists.txt
```diff
@@ -126,8 +126,9 @@ add_library(MediaServerLib
     ${CMAKE_CURRENT_LIST_DIR}/src/VideoLayerSelector.cpp
     ${CMAKE_CURRENT_LIST_DIR}/src/VideoCodecFactory.cpp
     ${CMAKE_CURRENT_LIST_DIR}/src/VideoDecoderWorker.cpp
-    ${CMAKE_CURRENT_LIST_DIR}/src/audiodecoder.cpp
-    ${CMAKE_CURRENT_LIST_DIR}/src/audiotransrater.cpp
+    ${CMAKE_CURRENT_LIST_DIR}/src/AudioDecoderWorker.cpp
```
murillo128, line 129: 🥳

include/AudioPipe.h
```diff
@@ -15,14 +17,14 @@ class AudioPipe :
 	virtual ~AudioPipe();
 	
 	//Audio input
-	virtual int RecBuffer(SWORD *buffer,DWORD size);
+	virtual DWORD RecBuffer(DWORD frameSize, const AudioBuffer::shared& audioBuffer);
```
murillo128, line 20: Let's return it instead
```suggestion
	virtual  AudioBuffer::shared RecBuffer(DWORD frameSize);
```

include/AudioPipe.h
```diff
@@ -15,14 +17,14 @@ class AudioPipe :
...
 	virtual int ClearBuffer();
 	virtual void CancelRecBuffer();
 	virtual int StartRecording(DWORD rate);
 	virtual int StopRecording();
 
 	//Audio output
-	virtual int PlayBuffer(SWORD *buffer,DWORD size,DWORD frameTime, BYTE vadLevel = -1);
+	virtual int PlayBuffer(SWORD *buffer,DWORD size, QWORD pts, BYTE vadLevel = -1);
```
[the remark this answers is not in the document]
yezhan10: My initial thought is to push audio buffer to a queue. I discarded this since decoded audio frame size and required audio frame size may differ, for aac, each playbuffer we write 1024*numChannels worth of samples, but opus requires totally different frame size (960, 1920 etc) to encode. so the logic of consuming the pcm data might be complicated, and i still need a small fifo buffer serves as a cache in case requested number of samples from encoder is less than number of sampling in audio buffer.

include/AudioPipe.h
```diff
@@ -38,6 +40,14 @@ class AudioPipe :
 
 	//Members
 	fifo<SWORD,48000*4>	fifoBuffer;
+	std::queue<std::pair<QWORD, QWORD>> audioPts;
```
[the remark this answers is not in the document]
yezhan10: i think it might be more complicated if we have a queue of audio buffers? I will give more thoughts on this..

src/AudioPipe.cpp
```diff
@@ -262,33 +212,156 @@ int AudioPipe::RecBuffer(SWORD* buffer, DWORD size)
...
+		if(!audioPts.empty() && ref > audioPts.front().first) 
+		{
+			reset = 1;
+			break;
+		}
+	}
+	// if timestamp goes backwards
+	if(reset) 
+	{
+		if(encoderPTS + frameSize <= ref + decoderFrameSize)
```
bcostdolby, line 244: Please rename everything to be consistent. Its difficult to read as every time i find a new variable I need to trace back how it is used to know what it is.

I think (are these meanings correct?):
* encoderNextPTS = encoderPTS
* encoderCurrentPTS = ts
* encodeFrameSize = frameSize 
* decoderNextPTS = ref
* decoderFrameSize = decoderFrameSize
* decoderFirst = first

src/AudioPipe.cpp
```diff
@@ -262,33 +212,156 @@ int AudioPipe::RecBuffer(SWORD* buffer, DWORD size)
...
+			ts = encoderPTS;
+		else
+		{
+			// no need to remove any data from fifoBuffer in case of time wraparound
+			// since timestamps are considered as continuous
```
bcostdolby, line 249: But didnt we just maybe drop multiple items from the audioPts queue above?

src/AudioPipe.cpp
bcostdolby: I may have been very wrong in the above...

It looks like encoderPTS at the start of this function call holds  the PTS expected to be given to the first sample in the next block we pull from the fifo?

src/AudioPipe.cpp
```diff
@@ -262,33 +212,156 @@ int AudioPipe::RecBuffer(SWORD* buffer, DWORD size)
...
+	// ref is found and perform ts interpolation
+	else if(ref != -1)
+	{
+		// if we have two decoder pts for ts interpolation
+		if(!audioPts.empty())
+		{
+			// if decoder timestamp jump more than 2 frames
+			if(audioPts.front().first - ref > 2*decoderFrameSize)
+			{
+				if(encoderPTS + frameSize <= ref + decoderFrameSize)
```
bcostdolby, line 264: Think I need to ask about this as well

murillo128:
In order to be able to progress this PR, can we split it in two PRs? 

The first one about method changes to use the AudioBuffer and the second one to fix the timestamp calculation inside the AudioPipe?

yezhan10:
> In order to be able to progress this PR, can we split it in two PRs?
> 
> The first one about method changes to use the AudioBuffer and the second one to fix the timestamp calculation inside the AudioPipe?

@murillo128 @bcostdolby I have updated the implementation to use a queue of audio buffers in audio pipe, and also added timestamp calculation for audio buffer based on decoder pts.  if we all like this approach better, i will clean up the code a bit in my next commit. This version still has `Encode(const AudioBuffer::const_shared& audioBuffer, BYTE* out, int outLen)`, I will update this to return an encoded frame in next PR most likely. Also, maybe we can use audiobuffer pool as an improvement in the next PR as well to avoid frequent audio buffer creation.

include/AudioBuffer.h
```diff
@@ -0,0 +1,109 @@
...
+		memcpy((SWORD*)pcmBuffer.data(), in, sizeof(SWORD)*totalResampled);
+		return numSamples;
+	}
+
+	int CopyDecodedData(uint8_t** pcmData, uint32_t numSamples)
+	{
+		if(!pcmData || !*pcmData)
+			return Error("-AudioBuffer::CopyDecodedData() invalid frame data pointer\n");
+	
+		if(numSamples*numChannels > pcmBuffer.capacity())
```
bcostdolby, line 79: Probably should be looking at size not capacity unless we are resizing the buffer as we go.

Also the each channel is doing a std::min with 2 and this does not so will miss the check

include/AudioBuffer.h
```diff
@@ -0,0 +1,109 @@
...
+			pcmBuffer.resize(totalResampled);
+		}
+			
+		memcpy((SWORD*)pcmBuffer.data(), in, sizeof(SWORD)*totalResampled);
+		return numSamples;
+	}
+
+	int CopyDecodedData(uint8_t** pcmData, uint32_t numSamples)
+	{
+		if(!pcmData || !*pcmData)
```
bcostdolby, line 76: Possibly need to check more than just pcmData[0] != nullptr but also for all other possible channels. 

Overall this API isnt very safe might be worth rethinking. We have num samples passed in but not num channels and it is implicitly expected to be equal to min(numChannels, 2) which is non obvious

src/AudioPipe.cpp
```diff
@@ -156,78 +154,65 @@ int AudioPipe::StopPlaying()
...
-
-	//if not enought
-	if (totalSize > left)
-	{
-		Warning("-AudioPipe::PlayBuffer() | not enought space %d %d\n", fifoBuffer.size(), fifoBuffer.length());
-		//Free space
-		fifoBuffer.remove(totalSize - static_cast<DWORD>(left / numChannels)* numChannels);
+		QWORD decoderPTS = audioBuffer->GetTimestamp();
+		// scale the decoder pts based on record rate
+		QWORD scaledDecoderPTS = std::round<QWORD>((decoderPTS-decoderPTSOffset) * (recordRate / (double)playRate)) + decoderPTSOffset;
```
bcostdolby, line 191: Tomorrow when we chat again, might need to understand why we need to scale the PTS based on the resampling. I.e. For example if the ingest clock rate as from SRT is say 1000Hz and the PTS is 5000 (so 5 sec in roughly) and the incoming audio sample rate is say 16KHz (which is different from the PTS clock).

Lets say we need to resample to 32KHz we still would push PTS meta-data for the first sample in the block with a PTS = 5000 and a PTS Clock = 1000Hz as although there are now twice as many samples, they are still for the same point in time according to the PTS clock.

I dont see us changing the PTS clock in the code here (dont think we need to, but to work with this rescaling of PTS I assume we would also need to rescale the PTS clock from 1000Hz to 2000Hz).

src/AudioPipe.cpp
```diff
@@ -239,21 +224,21 @@ int AudioPipe::RecBuffer(SWORD* buffer, DWORD size)
...
 	}
-	
-	//Calculate total audio length
-	totalSize = size * numChannels;
-	
-	//Until we have enought samples
-	while (!canceled && recording && (fifoBuffer.length() < totalSize + cache))
+	// TODO: use object pool to reuse audiobuffer
+	AudioBuffer::shared audioBuffer = std::make_shared<AudioBuffer>(frameSize, numChannels);
+	audioBuffer->SetClockRate(recordRate);
```
bcostdolby, line 235: Ahh related to comment above. I see we are changing the PTS clock here. I dont think we need to do this right?

I.e. The sample rate clock and the PTS clock are two different things arent they?

src/AudioPipe.cpp
```diff
@@ -262,33 +247,117 @@ int AudioPipe::RecBuffer(SWORD* buffer, DWORD size)
...
 
-	//Get samples from queue
-	len = fifoBuffer.pop(buffer, totalSize) / numChannels;
+	AudioBuffer::const_shared frontAudioBuffer;
+	int cachedSamples = rateBlockBuffer.length() / numChannels;
+	// read from rate block buffer first
+	// if rateBlockBuffer has enough data, no need to pop the audio buffer queue
+	if(cachedSamples >= frameSize)
+	{
+		rateBlockBuffer.pop(audioBuffer->GetAudioBufferData(requiredNumSamples), requiredNumSamples);
```
bcostdolby, line 261: Dont need separate if statement block for this. Just take min(requiredNumSamples,cachedSamples) from cache (might be 0 or more up to needed size)

I.e. I think most of these if/else blocks can be flattened into a single loop.

include/AudioBuffer.h
```diff
@@ -0,0 +1,109 @@
...
+		int totalSamples = numChannels * numSamples;
+		if(totalSamples > pcmBuffer.capacity())
+			return Error("AudioBuffer::CopyAudioBufferData() exceed audio buffer capacity, want=%d, capacity=%d\n", totalSamples, pcmBuffer.capacity());
+		memcpy(this->GetAudioBufferData(totalSamples), in, totalSamples*sizeof(SWORD));
+		return totalSamples;
+	}
+
+	void Resize(int numSamples) 
+	{
+		pcmBuffer.resize(numSamples*numChannels);
```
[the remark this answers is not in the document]
yezhan10: i removed this function since it's unused

include/AudioBuffer.h
yezhan10: the min(numChannels,2) is from the old code, i guess this indicates that our system does not support multichannel audio.  i have removed the min(numChannels,2) and added check for each channel's data pointer

include/AudioBuffer.h
```diff
@@ -0,0 +1,114 @@
...
+					pcmBuffer[i * numChannels + ch] = ((float*)(pcmData[ch]))[i] * (1 << 15);
+				else 
+					return Error("-AudioBuffer::CopyDecodedData() invalid data pointer for ch %d\n", ch);
+				totalWrittenSamples++;
+			}
+
+		}
+		return totalWrittenSamples / numChannels;		
+	}
+	bool AddPCM(SWORD val, int pos)
```
murillo128, line 94: I am not sure if I prefer having a getter for the `std::vector` or overloading the  `[]` instead. 

At most
```suggestion
	bool SetSample(size_t pos, SWORD sample) 
```

include/AudioBuffer.h
```diff
@@ -0,0 +1,114 @@
...
+	const int GetNumChannels() const { return numChannels; }
+	const int GetNumSamples() const { return pcmBuffer.size() / numChannels; }
+
+private:
+	int numSamplesPerFrame;
+	int numChannels;	
+	QWORD ts = 0;
+	DWORD clockRate = 0;
+	std::vector<SWORD> pcmBuffer;
+	SWORD* data;
```
murillo128, line 110: I don't think we should keep reader state inside this class

include/AudioPipe.h
```diff
@@ -34,10 +37,11 @@ class AudioPipe :
 private:
 	//Los mutex y condiciones
 	pthread_mutex_t mutex;
-	pthread_cond_t  cond; 
-
+	pthread_cond_t  cond;
 	//Members
-	fifo<SWORD,48000*4>	fifoBuffer;
+	fifo<SWORD,48000*4>	rateBlockBuffer;
+	QWORD encoderPTS = 0;
```
murillo128, line 43: let's remove for now

src/opus/OpusEncoder.h
```diff
@@ -16,13 +17,16 @@ class OpusEncoder : public AudioEncoder
 public:
 	OpusEncoder(const Properties &properties);
 	virtual ~OpusEncoder();
-	virtual int Encode(SWORD *in,int inLen,BYTE* out,int outLen);
+	virtual AudioFrame* Encode(const AudioBuffer::const_shared& audioBuffer);
```
murillo128, line 20: can we return a AudioFrame::shared better? We could add an <ObjectPool> to the  encoders for increasing performance if needed.

include/audio.h
```diff
@@ -83,7 +84,7 @@ class AudioEncoder
 public:
 	// Must have virtual destructor to ensure child class's destructor is called
 	virtual ~AudioEncoder(){};
-	virtual int   Encode(SWORD *in,int inLen,BYTE* out,int outLen)=0;
+	virtual AudioFrame* Encode(const AudioBuffer::const_shared& audioBuffer) = 0;
```
murillo128, line 87: can we use a shared pointer better?### 
```suggestion
	virtual AudioFrame::shared Encode(const AudioBuffer::const_shared& audioBuffer) = 0;

```

murillo128 (changes requested):
looking better, still some changes I would like to make

include/AudioBuffer.h
```diff
@@ -0,0 +1,114 @@
...
+		
+	}
+	QWORD	GetTimestamp() const	{ return ts; }
+	void	SetTimestamp(QWORD ts)	{ this->ts = ts; }
+	
+	QWORD	GetClockRate() const	{ return clockRate; }
+	void	SetClockRate(DWORD clockRate) { this->clockRate = clockRate; }
+
+	const SWORD* GetData() const { return pcmBuffer.data(); }
+	SWORD* ConsumeData(int numSamples)
```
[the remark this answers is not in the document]
yezhan10: this function is used in gsmcodec.cpp gsm_decode() line 92/96.

include/AudioBuffer.h
yezhan10: I chose to use getter function and i would rename it to SetSampleAt() to have more obvious distinction between SetSamples()

include/AudioBuffer.h
yezhan10: the reader position is needed in ConsumeSamples() function which is used in gsmcodec.cpp. I will leave it for now.

include/AudioBuffer.h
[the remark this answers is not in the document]
murillo128: I don't like keeping a read state inside the `AudioBuffer`. The same `AudioBuffer` could be passed to multiple `AudioEncoders` so it feels weird do modify the state just when reading samples.

src/gsm/gsmcodec.cpp
```diff
@@ -51,42 +55,49 @@ GSMDecoder::~GSMDecoder()
...
-		//GSM de M$ vienen 2 paquetes seguidos
-		if (outLen<160*2)
-			return 0;
-
+		int outLen = 160*2;
+		auto audioBuffer = std::make_shared<AudioBuffer>(outLen, 1);
 		//Decodificamos el primero
-		if (gsm_decode(g,(gsm_byte *)in,(gsm_signal *)out)<0)
-			return 0;
+		if (gsm_decode(g,(gsm_byte *)in,(gsm_signal *)audioBuffer->ConsumeSamples(160))<0)
```
murillo128, line 93:
```suggestion
		if (gsm_decode(g,(gsm_byte *)in,(gsm_signal *)audioBuffer->GetData(160))<0)
```

src/gsm/gsmcodec.cpp
```diff
@@ -51,42 +55,49 @@ GSMDecoder::~GSMDecoder()
...
+			return {};
 
 		//Y el segundo
-		if (gsm_decode(g,(gsm_byte *)&in[33],(gsm_signal *)&out[160])<0)
-			return 0;
+		if (gsm_decode(g,(gsm_byte *)&in[33],(gsm_signal *)audioBuffer->ConsumeSamples(160))<0)
```
murillo128, line 97:
```suggestion
		if (gsm_decode(g,(gsm_byte *)&in[33],(gsm_signal *)(audioBuffer->GetData()+160))<0)
```

src/gsm/gsmcodec.cpp
```diff
@@ -51,42 +55,49 @@ GSMDecoder::~GSMDecoder()
...
-int GSMDecoder::Decode(const AudioFrame::const_shared& audioFrame, SWORD* out,int outLen)
+int GSMDecoder::Decode(const AudioFrame::const_shared& audioFrame)
 {
 	//Dependiendo de la longitud tenemos un tipo u otro
 	int inLen = audioFrame->GetLength();
-	const uint8_t* in = audioFrame->GetData();
+	uint8_t* in =  const_cast<uint8_t*>(audioFrame->GetData());
+	if(!in || inLen==0) 
+		return 0;
+	audioFrameInfo.first = in;
```
murillo128, line 65: why don't acually decode here and have a FIFO of the produced `AudioBuffers` that are consumed in the `GetDecodedAudioFrame()`?

As you are not keeping the `audioFrame` shared pointer, `in` pointer can be invalid when `GetDecodedAudioFrame()` is called

murillo128 (changes requested):

src/gsm/gsmcodec.cpp
yezhan10: i agree that decode here instead of in GetDecodeAudioFrame() is more safe. Do we need a queue of audiobuffers stored for these codecs(g711 etc)? Looks like audioFrame only contains one decoded frame for these codecs.

src/g711/pcmacodec.cpp
```diff
@@ -25,31 +26,49 @@ PCMADecoder::~PCMADecoder()
...
+		else	
+			return 0;
+	}
+	audioBufferQueue.push(std::move(audioBuffer));
+	return 1;	
 }
+
+AudioBuffer::shared PCMADecoder::GetDecodedAudioFrame()
+{
+	if(!audioBufferQueue.empty())
```
murillo128, line 67: not big deal, but I preffer to fail early

```
if(audioBufferQueue.empty())
  return {};
auto audioBuffer = audioBufferQueue.front();
audioBufferQueue.pop();
return audioBuffer;
}

```

murillo128 (approved):
lgtm

PR #336, opened 2024-09-04
A single function to get CompositionTimeOffset for both RTMP and eRTMP

harryz2000:
This will fix the hevc pts issue.

### DIFF
--- a/src/rtmp/rtmpmessage.cpp
+++ b/src/rtmp/rtmpmessage.cpp
@@ -955,7 +955,7 @@ void RTMPVideoFrame::Dump()
 		
 		if (codecEx==VideoCodecEx::HEVC && packetType == CodedFrames)
 		{
-			Debug("\t[HEVC composition time: %d/]\n", GetSI24(&extraData[1]));
+			Debug("\t[HEVC composition time: %d/]\n", GetSI24(extraData));
 		}
 			
 		if (codecEx==VideoCodecEx::HEVC && IsConfig())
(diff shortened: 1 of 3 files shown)


bcostdolby:
Maybe remove the Get/Set AVCTS methods and replace with GetCTS and SetCTS that are not AVC specific but based on if it is extended or not will set/get the correct byte of the extra data. 

Right now we are saying it is extended and then modifying the meaning of extra data.

src/rtmp/rtmpmessage.cpp
```diff
@@ -978,6 +978,21 @@ void RTMPVideoFrame::Dump()
 	}
 }
 
+int32_t RTMPVideoFrame::GetCompositionTime() const
+{
+	if (isExtended)
+	{
+		if (GetVideoCodecEx() == VideoCodecEx::HEVC || GetVideoCodecEx() == VideoCodecEx::H264)
+			return  GetSI24(extraData);
```
bcostdolby, line 986: Is this correct as a SI24? I assume that means signed? The original CTS was a unsigned

harryz2000: it is signed yes.

SI24                           ¦Signed 24-bit integer

harryz2000: ¦      // See ISO/IEC 14496-12, 8.15.3 for an explanation of composition times.                                                                                           ¦
¦      // The offset in an FLV file is always in milliseconds.                                                                                                            ¦
¦      compositionTimeOffset = SI24                                                                                                                                       ¦
¦

harryz2000:
> Maybe remove the Get/Set AVCTS methods and replace with GetCTS and SetCTS that are not AVC specific but based on if it is extended or not will set/get the correct byte of the extra data.
> 
> Right now we are saying it is extended and then modifying the meaning of extra data.

Added @todo. We should remove the AVC functions. yes.

harryz2000:
> Maybe remove the Get/Set AVCTS methods and replace with GetCTS and SetCTS that are not AVC specific but based on if it is extended or not will set/get the correct byte of the extra data.
> 
> Right now we are saying it is extended and then modifying the meaning of extra data.

Added todo to remove AVC specific functions. Renamed function to GetComponsitionTime().

bcostdolby (approved):
lgtm

murillo128 (approved):
lgtm

PR #340, opened 2024-09-09
refactor: use BufferReader within BitReader

murillo128:
This is the first step to remove the memory copy when doing RBSP unescaping on h26x codecs.

For doing so, this PR changes the internal data of the `BitReader` to use a `BufferReader`. On the short term, it allows us to no have to specify the amount of bites we want to read at the `BitReader` and it will be consuming from the `BufferReader` instead.

For beins able to handle the RBSP unscapeing, we will have to make a `BufferReader`  that performas internally the rbsp unscaping and just returns the proper values to the `GetX` and move the pointer internally instead.

The only drawback of this PR, is that in order to continue using the `BufferReader` after passing it to a `BitReader`, we need to do `Flush()` to move back unused cache back to the reader.

We may improve the behaviour of the `BitReader` to not do eager caching and only cache the bits needed instead of reading always 32, or passing the `BufferReader` by move to the `BitReader` and then regain it at the `Flush()` method to try to enforce calling `Flush()`


mildsunrise (approved):
provisional lgtm from me, I'll review more deeply later.
i'm neutral when it comes to this change, since it looks to me like it only helps in the case where bit data and byte data are intermixed together, but in most cases I think it is clearly delimited which sections contain bit data and which don't.

the fact that BitReader consumes eagerly from the BufferReader and you have to call Flush() looks a biiit error-prone to me but it's okay, i don't have anything better to suggest anyway

murillo128:
We will have to change the `BitReader` so it only consumes input when Get()/Skip() goes over byte boundaries, I will do it in an upcoming PR.

After that the idea is to change the `BitReader` to be a template receiving a `BufferReader` like object

```
template< class T = BufferReader>
class BitReader
{
}
```

And then create a `RBSPEcapedBufferReader` object that implements the method required by the `BitReader` escaping the h26x bitstreams

PR #341, opened 2024-09-10
Avoid TimeServiceWrapper misuse

harryz2000:
With this change, the object has to be created through Create() function, which would be shared_ptr. As shared_from_this() wouldn't work if the object is not created through shared pointer way. This change would help to avoid such misuse.

An example is provided.

### DIFF
--- a/include/FrameDelayCalculator.h
+++ b/include/FrameDelayCalculator.h
@@ -9,11 +9,11 @@
 #include <queue>
 #include <algorithm>
 
-class FrameDelayCalculator : public std::enable_shared_from_this<FrameDelayCalculator>
+class FrameDelayCalculator : public TimeServiceWrapper<FrameDelayCalculator>
 {
 public:
 
-	FrameDelayCalculator(int aUpdateRefsPacketLateThresholdMs, std::chrono::milliseconds aUpdateRefsStepPacketEarlyMs, TimeService& timeService);
+	FrameDelayCalculator(Protected prt, int aUpdateRefsPacketLateThresholdMs, std::chrono::milliseconds aUpdateRefsStepPacketEarlyMs, TimeService& timeService);
 	
 	/**
 	 * Calculate the dispatch delay for the arrived frame. This function is thread safe.
--- a/include/FrameDelayCalculator.h
+++ b/include/FrameDelayCalculator.h
@@ -12,8 +12,6 @@
 class FrameDelayCalculator : public TimeServiceWrapper<FrameDelayCalculator>
 {
 public:
-
-	FrameDelayCalculator(Protected prt, int aUpdateRefsPacketLateThresholdMs, std::chrono::milliseconds aUpdateRefsStepPacketEarlyMs, TimeService& timeService);
 	
 	/**
 	 * Calculate the dispatch delay for the arrived frame. This function is thread safe.
@@ -54,6 +52,8 @@ private:
 		
 		__uint128_t field = 0;
 	};
+
+	FrameDelayCalculator(int aUpdateRefsPacketLateThresholdMs, std::chrono::milliseconds aUpdateRefsStepPacketEarlyMs, TimeService& timeService);
 	
 	/**
 	 * Get the delay of current frame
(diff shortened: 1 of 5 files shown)


murillo128:
how about making the constructor private and using `friend`?

```
#include <memory>
#include <utility>

template <typename T>
class TimeServiceWrapper : public std::enable_shared_from_this<T>
{
public:
	template <typename... ARGS>
	static std::shared_ptr<T> Create(ARGS&&... args)
	{
		return std::shared_ptr<T>(new T(std::forward<ARGS>(args)...));
	}
};

class X : public TimeServiceWrapper<X>
{
private:
    X(){}

   friend class TimeServiceWrapper<X>;    
};


int main()
{
    auto x = X::Create();
}
```

https://godbolt.org/z/c4v8Eo3Ts

(note: i was not able to make `std::make_shared` work, so had to use `new` instead)

murillo128 (approved):
using factories lgtm, the only concern is the usage of the `Protected` which seems ackward and non intuitive

harryz2000:
> how about making the constructor private and using `friend`?
> 
> ```
> #include <memory>
> #include <utility>
> 
> template <typename T>
> class TimeServiceWrapper : public std::enable_shared_from_this<T>
> {
> public:
> 	template <typename... ARGS>
> 	static std::shared_ptr<T> Create(ARGS&&... args)
> 	{
> 		return std::shared_ptr<T>(new T(std::forward<ARGS>(args)...));
> 	}
> };
> 
> class X : public TimeServiceWrapper<X>
> {
> private:
>     X(){}
> 
>    friend class TimeServiceWrapper<X>;    
> };
> 
> 
> int main()
> {
>     auto x = X::Create();
> }
> ```
> 
> https://godbolt.org/z/c4v8Eo3Ts
> 
> (note: i was not able to make `std::make_shared` work, so had to use `new` instead)

I also tried similar way. But didn't work. Maybe I was using make_shared(). will try it tommorow.

harryz2000:
I changed to use private constructor. Also added timeService as a member.

harryz2000:
> > how about making the constructor private and using `friend`?
> > ```
> > #include <memory>
> > #include <utility>
> > 
> > template <typename T>
> > class TimeServiceWrapper : public std::enable_shared_from_this<T>
> > {
> > public:
> > 	template <typename... ARGS>
> > 	static std::shared_ptr<T> Create(ARGS&&... args)
> > 	{
> > 		return std::shared_ptr<T>(new T(std::forward<ARGS>(args)...));
> > 	}
> > };
> > 
> > class X : public TimeServiceWrapper<X>
> > {
> > private:
> >     X(){}
> > 
> >    friend class TimeServiceWrapper<X>;    
> > };
> > 
> > 
> > int main()
> > {
> >     auto x = X::Create();
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > https://godbolt.org/z/c4v8Eo3Ts
> > (note: i was not able to make `std::make_shared` work, so had to use `new` instead)
> 
> I also tried similar way. But didn't work. Maybe I was using make_shared(). will try it tommorow.



> > how about making the constructor private and using `friend`?
> > ```
> > #include <memory>
> > #include <utility>
> > 
> > template <typename T>
> > class TimeServiceWrapper : public std::enable_shared_from_this<T>
> > {
> > public:
> > 	template <typename... ARGS>
> > 	static std::shared_ptr<T> Create(ARGS&&... args)
> > 	{
> > 		return std::shared_ptr<T>(new T(std::forward<ARGS>(args)...));
> > 	}
> > };
> > 
> > class X : public TimeServiceWrapper<X>
> > {
> > private:
> >     X(){}
> > 
> >    friend class TimeServiceWrapper<X>;    
> > };
> > 
> > 
> > int main()
> > {
> >     auto x = X::Create();
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > https://godbolt.org/z/c4v8Eo3Ts
> > (note: i was not able to make `std::make_shared` work, so had to use `new` instead)
> 
> I also tried similar way. But didn't work. Maybe I was using make_shared(). will try it tommorow.

Yes. make_shared() doesn't work. I am fine to use new operator.

PR #342, opened 2024-09-12
refactor: [DIOS-6937] Use safer async wrapper with weak_ptr object lifetimes

murillo128:


include/TimeService.h
```diff
@@ -52,6 +52,12 @@ class TimeServiceWrapper : public std::enable_shared_from_this<T>
 	TimeServiceWrapper(TimeService& timeService) : timeService(timeService)
 	{
 	}
+
+
+	inline void Sync(const std::function<void(std::chrono::milliseconds)>& func)
```
harryz2000, line 57: I would prefer to use template, which is faster.

include/TimeService.h
```diff
@@ -63,6 +69,17 @@ class TimeServiceWrapper : public std::enable_shared_from_this<T>
 			func(self, now);
 		});
 	}
+
+	template <typename Func>
+	void AsyncSafe(Func&& func, const std::function<void(std::chrono::milliseconds)>& callback)
```
harryz2000, line 74: same as above

harryz2000:
When I search the Async use by use `.Async(` , it looks there are still some code might needs to be changed to use the wrapper:

```
dev/media-server/src/rtp/RTPIncomingMediaStreamDepacketizer.cpp
  94,13: 	timeService.Async([=](auto now){

dev/media-server/src/DTLSICETransport.cpp
  2255,13: 	timeService.Async([=](auto now) {
  3299,13: 	timeService.Async([=](auto now) {

dev/media-server/src/mp4recorder.cpp
  727,6: 	loop.Async([=](auto now){
  749,6: 	loop.Async([=](auto now){
  825,6: 	loop.Async([=,cloned = frame.Clone()](auto now){

dev/media-server/src/RTPBundleTransport.cpp
  83,6: 	loop.Async([=, fd = std::move(fd)](std::chrono::milliseconds) {
  91,6: 	loop.Async([=](std::chrono::milliseconds) { 
  172,6: 	loop.Async([=](auto now){
  189,6: 	loop.Async([=](auto now){
  247,6: 	loop.Async([=](auto now) {
  727,6: 	loop.Async([=](auto now){
  751,6: 	loop.Async([=](auto now){
```

src/ActiveSpeakerMultiplexer.cpp
```diff
@@ -156,28 +156,27 @@ void ActiveSpeakerMultiplexer::onRTP(const RTPIncomingMediaStream* incoming, con
 		//Exit
 		return;
 
-	timeService.Async([=, packet = packet](auto) {
+	AsyncSafe([&, packet = packet](auto self, auto now) {
```
harryz2000, line 159: Can use change to capture = rather than &, it is dangerous if it is capturing a local variable.

src/ActiveSpeakerMultiplexer.cpp
```diff
@@ -215,14 +214,14 @@ void ActiveSpeakerMultiplexer::onEnded(const RTPIncomingMediaStream* incoming)
 	if (!incoming)
 		return;
 
-	timeService.Async([=](auto) {
+	AsyncSafe([&](auto self, auto now) {
```
harryz2000, line 217: Same as above. I think capturing & is dangerous if it accidentally capture a local variable although have the protection for this object.

src/dtls.cpp
```diff
@@ -713,13 +713,13 @@ int DTLSConnection::Renegotiate()
 {
 	TRACE_EVENT("dtls", "DTLSConnection::Renegotiate");
 	//Run in event loop thread
-	timeService.Async([this](auto now){
+	AsyncSafe([&](auto self, auto now){
```
harryz2000, line 716: Consider remove & and use self->ssl at line 717

include/TimeService.h
harryz2000: done.

include/TimeService.h
harryz2000: done.

bcostdolby:
> When I search the Async use by use `.Async(` , it looks there are still some code might needs to be changed to use the wrapper:

I am in the process of renaming the original function to AsyncUnsafe (and other functions) and seeing what errors the compiler comes up with.

src/ActiveSpeakerMultiplexer.cpp
bcostdolby: I will get around to reviewing the captures after everything else is updated and building + running. I think I might change them to not sure & or = but specify what they need if the list isn't too long as it will prevent using this-> (which is safe, but we want to use self-> apparently)

murillo128 (commented):
lgtm

PR #344, opened 2024-09-13
Use exception rather than assertion for parsing data

harryz2000:


include/bitstream/BitReader.h
```diff
@@ -39,7 +39,8 @@ class BitReader
 
 	inline DWORD Get(DWORD n)
 	{
-		assert(n <= 32);
+		if (n > 32)
```
bcostdolby, line 42: I prefer this, but it means that someone is calling Get() asking for more bits of data than are possible to fit in the DWORD. I.e. Is a bug in the caller really which assert() is theoretically correct for.

harryz2000: I don't think we should be that strict. Passing n shouldn't be illegal but just invalid. The argument could be from data not from our manual input. So I don't think calling it with a value more than 32 is a bug. If we use assert, we must check it before passing it to this Get() function, which would be not convenient and I think that is against our intention to use exception.

bcostdolby (approved):
lgtm

PR #346, opened 2024-09-16
feat: introduce rbsp unscaping buffer reader

murillo128:
completely untested, ptal

### DIFF
--- a/src/h264/H26xNal.h
+++ b/src/h264/H26xNal.h
@@ -114,7 +114,7 @@ protected:
 			else 
 			{
 				//Acu value
-				value = value << 1 | data[i];
+				value = value << 8 | data[i];
 				//Check consecutive zeros
 				if (data[i] == 0)
 					z++;
--- a/src/h264/H26xNal.h
+++ b/src/h264/H26xNal.h
@@ -6,15 +6,137 @@
 #include <optional>
 #include <functional>
 
+
 #include "config.h"
 #include "log.h"
 #include "tools.h"
 #include "Buffer.h"
 #include "BufferReader.h"
+#include "bitstream/BitReader.h"
 
 constexpr uint32_t AnnexBStartCode = 0x01;
 
 // H.264 NAL logic that can be shared with H.265 (mostly emulation prevention, annex B stream)
+
+class RbspReader
+{
+public:
+	RbspReader(const uint8_t* data, const size_t size) :
+		data(data),
+		size(size)
+	{
+		this->pos = 0;
+	}
+
+	inline bool   Assert(size_t num) const
+	{
+		DWORD len = 0;
+		DWORD i = pos;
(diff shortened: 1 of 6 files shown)


src/h264/H26xNal.h
```diff
@@ -6,16 +6,138 @@
...
 #include "tools.h"
 #include "Buffer.h"
 #include "BufferReader.h"
+#include "bitstream/BitReader.h"
 
 constexpr uint32_t AnnexBStartCode = 0x01;
 
 // H.264 NAL logic that can be shared with H.265 (mostly emulation prevention, annex B stream)
 
+class RbspReader
```
eck1001, line 21: It might be nice to add somewhere that Rbsp stands for `Raw Byte Sequence Payload`.

src/h264/H26xNal.h
```diff
@@ -6,16 +6,138 @@
...
+{
+public:
+	RbspReader(const uint8_t* data, const size_t size) :
+		data(data),
+		size(size)
+	{
+		this->pos = 0;
+	}
+
+	inline bool   Assert(size_t num) const
```
eck1001, line 31: What are we asserting here? 

Is this method used to skip the input NAL for a h26x buffer?

murillo128: this method is used to ensure that we have enough bytes to call PeekX method safely, we probably want to replace them with exceptions eventually. 

Note that this RbspReader as the same methods as the BufferReader so we can use them as template input for the BaseBitReader

src/h264/H26xNal.h
mildsunrise: I suggest documenting what this class does, something like

``` c
/**
 * This reader consumes an RBSP (usually the payload of an H.26x NALU)
 * and provides the unescaped bytes of the SODB, by skipping over
 * `emulation_prevention_three_byte`s as necessary (H.264 section 7.4.1.1 step 1).
 * Other codes (00 00 [00..02]) like start codes, as well as incomplete codes,
 * are currently passed through.
 */
```

src/h264/H26xNal.h
```diff
@@ -6,16 +6,138 @@
...
+			{
+				//Skip input
+				i++;
+				//No zeros
+				z = 0;
+			}
+			else 
+			{
+				//Acu value
+				value = value << 1 | data[i];
```
mildsunrise, line 117:
```suggestion
				value = value << 8 | data[i];
```

mildsunrise (approved):
yayyyy, finally ^^ thanks for implementing this.
other than the typo it lgtm. i'm sad we're copying the same bunch of code in three places though

murillo128:
thanxs for your reivews! yes, it is sad that I had to copy the same code in 3 places, couldn't find any good idea for cleaning it up, any idea on ow to improve it?

PR #348, opened 2024-09-23
refactor:calculate encoder pts based on decoder pts

yezhan10:


include/AudioBuffer.h
```diff
@@ -25,20 +25,21 @@ class AudioBuffer
 	void	 SetClockRate(DWORD clockRate) { this->clockRate = clockRate; }
 
 	const int16_t* GetData() const { return pcmBuffer.data(); }
-
-	uint16_t SetSamples(int16_t* in, uint16_t numSamples)
+	uint16_t SetSamples(int16_t* in, uint16_t numSamples, size_t offset=0, bool allowResize=false)
```
murillo128, line 28: not sure if I like the flag, is there any case in which do we do not want to allow resize?

if we want to allow resizing only for certain cases, wouldn't it be better to have an explicit `Resize()` method instead?

include/AudioBuffer.h
```diff
@@ -25,20 +25,21 @@ class AudioBuffer
...
+		if (allowResize && totalSamples != pcmBuffer.size())
 		{
-			Debug("AudioBuffer::SetSamples buffer resized, resized to =%d\n", totalResampled);
-			pcmBuffer.resize(totalResampled);
+			Debug("AudioBuffer::SetSamples buffer resized, resized to =%d\n", totalSamples);
+			pcmBuffer.resize(totalSamples);
 		}
-
-		memcpy((int16_t*)pcmBuffer.data(), in, sizeof(int16_t) * totalResampled);
+		else if (!allowResize && pcmBuffer.size() < offset+totalSamples)
```
murillo128, line 39: simplify the if/else please

include/AudioPipe.h
```diff
@@ -37,8 +38,16 @@ class AudioPipe :
...
 	pthread_cond_t  cond;
+	// track available num samples
+	QWORD availableNumSamples = 0;
+
+	CircularQueue<AudioBuffer::const_shared> queue;
 	//Members
 	fifo<SWORD,48000*4>	rateBlockBuffer;
+	QWORD rateBlockBufferPTS;
+
+	QWORD decoderFrameSize = 0;
```
murillo128, line 49: rename, there ir no `decoder` in this context so not sure what is referring too just by looking at the variable name

src/AudioPipe.cpp
```diff
@@ -168,6 +165,8 @@ int AudioPipe::PlayBuffer(const AudioBuffer::shared& audioBuffer)
 		//Ok
 		return size;
 	}
+	if (decoderPTSOffset == std::numeric_limits<QWORD>::max())
```
murillo128, line 168: better to use `std::optional` instead

src/AudioPipe.cpp
```diff
@@ -181,42 +180,31 @@ int AudioPipe::PlayBuffer(const AudioBuffer::shared& audioBuffer)
...
-	//Get left space
-	DWORD left = rateBlockBuffer.size() - rateBlockBuffer.length();
-
-	//if not enought
-	if (totalSize > left)
-	{
-		Warning("-AudioPipe::PlayBuffer() | not enough space %d %d\n", rateBlockBuffer.size(), rateBlockBuffer.length());
-		//Free space
-		rateBlockBuffer.remove(totalSize - static_cast<DWORD>(left / numChannels)* numChannels);
+		auto decoderPTS = audioBuffer->GetTimestamp();
```
murillo128, line 183: can we make `transrater` use `AudioBuffer` for in/out and do the pts calculus inside of it?

src/AudioPipe.cpp
```diff
@@ -254,17 +240,75 @@ AudioBuffer::shared AudioPipe::RecBuffer(DWORD frameSize)
...
+		// in case of pts discontinuity
+		else if(currDecoderPTS > prevPTS + 2*numSamples) 
+			break;
+		
+		prevPTS = currDecoderPTS;
+		numSamples = currDecoderSamples;
+
+		audioBuffer->SetSamples((SWORD*)frontAudioBuffer->GetData(), std::min(needSamples, currDecoderSamples), offset);
+		offset += std::min(needSamples, currDecoderSamples) * numChannels;
+		needSamples -= frontAudioBuffer->GetNumSamples();
```
murillo128, line 287: what happnes if `neededSampes<frontAudioBuffer->GetNumSamples()` wouldn't we be dropping samples?

src/AudioPipe.cpp
```diff
@@ -254,17 +240,75 @@ AudioBuffer::shared AudioPipe::RecBuffer(DWORD frameSize)
...
+	auto encoderPTS = bufferedSamples ? rateBlockBufferPTS : queue.front()->GetTimestamp();
+	// store prevPTS for pts wraparound/discontinuity detection
+	auto prevPTS = bufferedSamples ? rateBlockBufferPTS : queue.front()->GetTimestamp();;
+	// numSamples used to check ts discontinuity
+	auto numSamples = static_cast<int16_t>(bufferedSamples);
+	size_t offset = readSamples * numChannels;
+	while(queue.length()>0 && needSamples>0)
+	{
+		frontAudioBuffer = queue.front();
+		auto currDecoderPTS = frontAudioBuffer->GetTimestamp();
```
murillo128, line 265: imho it is simpler/safer to keep the timestamp of the first AudioBuffer as reference and not the latest one, it will simplify calculus

murillo128 (changes requested):

include/AudioBuffer.h
yezhan10: the idea is that the only allow resize in case of resampling. This SetSamples() is used in recBuffer() function as well where resize is disallowed.

murillo128: but resampling should produce a different audio buffer not reuse current one.  There should be possible to have multiple consumers for the same AudioBuffer, so the only ones allowed to change the AudioBiffers are the producers

src/AudioPipe.cpp
```diff
@@ -254,17 +240,75 @@ AudioBuffer::shared AudioPipe::RecBuffer(DWORD frameSize)
...
-			goto end;
+			pthread_mutex_unlock(&mutex);
+			return {};
 		}
 	}
-	//Get samples from queue
-	len = rateBlockBuffer.pop(buffer, totalSize) / numChannels;
-end:
-	//Unlock
+	AudioBuffer::const_shared frontAudioBuffer;
```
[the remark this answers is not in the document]
yezhan10: Since the audioBuffer created for encoder may include fractional number of audio buffer produced by decoders, I used rateBlockBuffer to store the leftover. For example, if encoder audio buffer requires 2.3 decoder audio buffers to make up the encoder frame size, this code block would pop 3 decoder audio buffers, and write the leftover( = 0.7*audio buffer samples) to rateBlockBuffer. To achieve this, it requires to read whatever in rateBlockBuffer first, if still need more samples, and then pop the audioBufferQueue based on number of samples still required from audio buffer queue.

src/AudioPipe.cpp
yezhan10: samples are not dropped, the leftover are stored in rateBlockBuffer.

src/AudioPipe.cpp
yezhan10: um i think the encoder PTS uses the first as reference: 
`auto encoderPTS = bufferedSamples ? rateBlockBufferPTS : queue.front()->GetTimestamp();`
if we have any samples in rateBlockBuffer, encoderPTS= rateBlockBufferPTS, otherwise, it equals to the first audio buffer. The pts related logic inside this while loop is mostly to infer rateBlockBufferPTS.

murillo128: why does it needs to be fine inside the loop? we just need to update it on the latest buffer, rigth?

include/AudioBuffer.h
yezhan10: maybe I can update the transrater to produce a new audio buffer instead of resizing?

src/AudioPipe.cpp
yezhan10: yeah most of the cases, rateBlockBufferPTS is updated based on the last buffer. And only in case of ts wraparound between rateBlockBuffer and first audio buffer, encoderPTS is updated inside.

murillo128 (commented):
lgtm so far, need to review the `RecBuffer()` though

src/AudioPipe.cpp
```diff
@@ -254,17 +232,85 @@ AudioBuffer::shared AudioPipe::RecBuffer(DWORD frameSize)
...
+			int diff = currPlayPTS-bufferedSamples;
+			// only need to update rateBlockBufferPTS if ts wraparound happened between rateBlockBuffer and audioBuffers.
+			if(diff>0 && prevPTS == rateBlockBufferPTS && bufferedSamples) 
+			{
+				rateBlockBufferPTS = diff;
+				recPTS = rateBlockBufferPTS;
+			}
+		}
+		// in case of pts discontinuity
+		else if (currPlayPTS > prevPTS + 2*numSamples)
```
murillo128, line 277: should we break in case of discontinuity?

include/AudioBufferPool.h
```diff
@@ -0,0 +1,91 @@
...
+	{
+		AudioBuffer* buffer;
+
+		//Get all the object from the pool
+		while (pool->try_dequeue(buffer))
+			//Delete them
+			delete(buffer);
+	}
+
+	AudioBuffer::shared Acquire()
```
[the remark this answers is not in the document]
yezhan10: I renamed to acquire since it seems to be more accurate. I will rename VideoBufferPool to use acquire instead.

src/AudioTransrater.cpp
```diff
@@ -58,20 +62,37 @@ void AudioTransrater::Close()
...
-	if (!out || !sizeOut)
-		//Error
-		return Error("-No output buffer/size [%p,%p]\n",out,sizeOut);
+	auto in = audioBuffer->GetData();
+	auto sizeIn = static_cast<uint32_t>(audioBuffer->GetNumSamples());
 
-	//Resample
-	int err = mcu_resampler_process_interleaved_int(resampler, (spx_int16_t*) in, (spx_uint32_t*) & sizeIn, (spx_int16_t*) out, (spx_uint32_t*) sizeOut);
+	auto bufferSize = static_cast<uint32_t>(std::ceil(static_cast<double>(outputRate) / inputRate * sizeIn));	
+	audioBufferPool.SetSize(bufferSize, audioBuffer->GetNumChannels());
```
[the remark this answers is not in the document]
yezhan10: according to the overall logic, num of channels in AudioDecoderWorker == number of channels in AudioPipe == num of channels in AudioTransrater. Unless there is channel count change happen in the middle of an audio stream, i would expect that audioBuffer->GetNumChannels always == numChannels in AudioTransrater. Probably put an assert? The audio would be corrupted if those two are not equal anyways unless we reinitialize the transrater with updated channel count.

src/AudioPipe.cpp
```diff
@@ -254,17 +232,85 @@ AudioBuffer::shared AudioPipe::RecBuffer(DWORD frameSize)
...
 	}
-	//Get samples from queue
-	len = rateBlockBuffer.pop(buffer, totalSize) / numChannels;
-end:
-	//Unlock
+	// since number of samples in decoded audio buffer and encoder frame size may differ,
+	// rateBlockBuffer is used to store the leftover. So always read number of samples
+	// from rateBlockBuffer first, and if more samples are needed, then read from queue.
+	AudioBuffer::const_shared frontAudioBuffer;
+	auto bufferedSamples = rateBlockBuffer.length() / numChannels;
```
murillo128, line 243: remove

src/AudioPipe.cpp
```diff
@@ -254,17 +232,85 @@ AudioBuffer::shared AudioPipe::RecBuffer(DWORD frameSize)
...
+	// readSamples: num samples per channel read from rateBlockBuffer
+	auto readSamples = std::min(bufferedSamples, frameSize);
+	// read whatever in rateBlockBuffer first
+	rateBlockBuffer.pop(buffer, readSamples*numChannels);
```
murillo128, line 247: too many variables here, makes it hard to follow and easy to make mistakes
```suggestion
	rateBlockBuffer.pop(audioBuffer->GetData(), readSamples);
```

src/AudioPipe.cpp
```diff
@@ -254,17 +232,85 @@ AudioBuffer::shared AudioPipe::RecBuffer(DWORD frameSize)
...
-	len = rateBlockBuffer.pop(buffer, totalSize) / numChannels;
-end:
-	//Unlock
+	// since number of samples in decoded audio buffer and encoder frame size may differ,
+	// rateBlockBuffer is used to store the leftover. So always read number of samples
+	// from rateBlockBuffer first, and if more samples are needed, then read from queue.
+	AudioBuffer::const_shared frontAudioBuffer;
+	auto bufferedSamples = rateBlockBuffer.length() / numChannels;
+	// readSamples: num samples per channel read from rateBlockBuffer
+	auto readSamples = std::min(bufferedSamples, frameSize);
```
murillo128, line 245:
```suggestion
	auto readSamples = std::min(rateBlockBuffer.length(), requiredNumSamples);
```

src/AudioPipe.cpp
```diff
@@ -254,17 +232,85 @@ AudioBuffer::shared AudioPipe::RecBuffer(DWORD frameSize)
...
+	// read whatever in rateBlockBuffer first
+	rateBlockBuffer.pop(buffer, readSamples*numChannels);
+	// update write offset since readSamples * numChannels already written into the buffer
+	size_t offset = readSamples * numChannels;
+	// needSamples: num samples per channel needed from queue after consuming rateBlockBuffer
+	int needSamples = frameSize - readSamples;
```
murillo128, line 251: reuse `requiredSamples`
```suggestion
	requiredSamples -= readSamples;
```

src/AudioPipe.cpp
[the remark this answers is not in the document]
murillo128: the code bellow is really hard to understand, too many variables and also mixes pts and frame read/pop, would it be possible to simplify it a bit further?

murillo128 (changes requested):

src/AudioPipe.cpp
yezhan10: remove the bufferedSamples variable? use rateBlockBuffer.length() / numChannels directly?

yezhan10: got it, have to continue to finish reading all comments.

src/AudioPipe.cpp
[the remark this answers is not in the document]
yezhan10: tried to remove unnecessary variables definitions. Sorry that i can not simplify the logic further

src/AudioPipe.cpp
```diff
@@ -254,17 +231,62 @@ AudioBuffer::shared AudioPipe::RecBuffer(DWORD frameSize)
...
+	auto readedSamples = std::min(rateBlockBuffer.length(), requiredNumSamples);
+	rateBlockBuffer.pop(const_cast<int16_t*>(audioBuffer->GetData()), readedSamples);
+	rateBlockBufferPTS += readedSamples;
+
+	while (queue.length()>0 && readedSamples<requiredNumSamples)
+	{
+		//Get reference to first buffer in queue 
+		auto& frontAudioBuffer = queue.front();
+		
+		if (frontAudioBuffer->GetTimestamp() > 2 * samples + prevPTS)
```
murillo128, line 259: shouldn't be the timestamp threshold for detecting a discontinuity be a static constant and not depending on the samples?

murillo128 (commented):
almost lgtm

src/AudioPipe.cpp
murillo128: yes, I meant a timestamp threshold converted to the num samples based on the output clock rate.

murillo128: what I meant is that we are detecting the disconituity based on the leftovers of the `rateBlockBuffer` which is variable

src/AudioTransrater.cpp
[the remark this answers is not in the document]
murillo128: `AudioDecoderWorker` may not be the only class using the transrater, we can't just ensure the behaviour of a class based on the use done from outside of it.

src/AudioPipe.cpp
```diff
@@ -254,17 +231,62 @@ AudioBuffer::shared AudioPipe::RecBuffer(DWORD frameSize)
...
+			break;
+		// Udpate variables for ts discontinuity check
+		prevPTS = frontAudioBuffer->GetTimestamp();
+		samples = frontAudioBuffer->GetNumSamples();
+
+		//The number of samples to read
+		auto totalSamplesInBuffer =  static_cast<int>(frontAudioBuffer->GetNumSamples() * numChannels);
+		auto neededSamples = std::min(totalSamplesInBuffer, requiredNumSamples - readedSamples);
+		// Read from queued buffer
+		audioBuffer->SetSamples(const_cast<int16_t*>(frontAudioBuffer->GetData()), neededSamples ,readedSamples);
```
murillo128, line 269: isn't it weird that `GetNumSamples` return the `samples` count  but `SetSamples` uses `samples * channel` ?

yezhan10: i will update the code so that GetNumSamples() return samples*channel.

murillo128 (approved):
lgtm

PR #349, opened 2024-09-24
DIOS-7112 Further refactor EventLoop to enable it to be used for different poll mechanism

harryz2000:


include/EventLoop.h
```diff
@@ -11,6 +11,154 @@
...
+	}
+};
+
+/**
+ * Concrete Poll implementation for Linux system poll mechanism.
+ */
+class SystemPoll : public Poll
+{
+public:
+	bool AddFd(FileDescriptor fd) override;
```
murillo128, line 139: we already have a file descriptor object , could we try to reuse it?

include/EventLoop.h
```diff
@@ -11,6 +11,154 @@
 
 using namespace std::chrono_literals;
 
+/**
```
murillo128, line 14: Move to it's own file

include/EventLoop.h
```diff
@@ -11,6 +11,154 @@
...
+		// calculate in that case. Assert here for reminder.
+		static_assert(sizeof(int) == 4);
+		return (uint64_t(static_cast<std::underlying_type<Poll::FileDescriptor::Category>::type>(fd.category)) << 32) + uint32_t(fd.fd);
+	}
+};
+
+/**
+ * Concrete Poll implementation for Linux system poll mechanism.
+ */
+class SystemPoll : public Poll
```
murillo128, line 136: Do we have any other kinf of Poll different than the SystemPoll?

harryz2000: Yes. I have created a SrtPoll for SRT server and client. It is in the PR in SRT repo.

include/EventLoop.h
harryz2000: I didn't realise we had a FileDescriptor class. Just checked that. It seems created for different purpose. The new FileDescriptor is only for wrap the information while the existing one is for some RATT feature. I would prefer to rename the new one to FileDescriptorInfo or similar to reflect it is very light struct to store some information, contain no logic.

include/Poll.h
```diff
@@ -0,0 +1,131 @@
...
+ * GetEvents()/GetError().
+ */
+class Poll
+{
+public:
+	
+	/**
+	 * Event types.
+	 */
+	enum Event
```
phil-moss-dolby, line 30: should we make this be an explicit type?

enum class Event : uint16_t {

harryz2000: That is a good point. I did think of change it to `enum class`, however that made it difficult to do bitwise operations. Such as:
`auto  events = Poll::Event::In | Poll::Event::Out;`
would become:
`auto events = static_cast<std::underlying_type<Poll::PollFd::Category>::type>(Poll::Event::In) | static_cast<std::underlying_type<Poll::PollFd::Category>::type>(Poll::Event::Out);`

It is very long and the readability looks not good to me. The reason as this enum is basically used as flags for bitwise calculation. It is a bit different to normal enumerations. We could change those to constants, but I feel Using a enum is to make code more organised. If the worries is the enum value could exceed 16bits, I can add a static_assert to it.

harryz2000: Actually I can specify uint16_t for a enum as well without use of class. Changed.

src/EventLoop.cpp
```diff
@@ -548,66 +554,98 @@ void EventLoop::Run(const std::chrono::milliseconds &duration)
...
+			}
+			
+			//Update now
+			now = Now();
+			
+			//UltraDebug("<EventLoop::Run() | poll timeout:%d timers:%d tasks:%d\n",timeout,timers.size(),tasks.size_approx());
+			
+			poll->ForEachFd([this](Poll::PollFd pfd) {
+
+				auto events = poll->GetEvents(pfd);
```
murillo128, line 587: why can't we return the error as an event?

src/EventLoop.cpp
```diff
@@ -548,66 +554,98 @@ void EventLoop::Run(const std::chrono::milliseconds &duration)
...
+
+				//Check error
+				auto error = poll->GetError(pfd);
+				if (error)
```
murillo128, line 591: I would prefer the nesting

```
switch (pdf.category)
{
  case Poll::PollFd::Category::Signaling:
  {
    if (error)
     .....
  }
  case Poll::PollFd::Category::IO
  {
  }
}

```

src/EventLoop.cpp
```diff
@@ -242,8 +260,16 @@ bool EventLoop::Start(int fd)
 		//Error
 		return Error("-EventLoop::Start() | could not start pipe [errno:%d]\n",errno);
 	
-	//Store socket
-	this->fd = fd;
+	
+	if (!poll->AddFd({Poll::PollFd::Category::Signaling, pipe[0]}))
```
murillo128, line 264: should the signaling pipe be managed inside the poll object itself?

src/rtmp/rtmpclientconnection.cpp
```diff
@@ -124,34 +128,31 @@ RTMPClientConnection::ErrorCode RTMPClientConnection::Connect(const char* server
...
+void RTMPClientConnection::OnLoopEnter()
 {
-	
-	EventLoop::Start([&]() {
-		run(this);
-	});
-	
-	return RTMPClientConnection::ErrorCode::NoError;
+	//Block signals to avoid exiting on SIGUSR1
+	blocksignals();
```
murillo128, line 142: should this be done inside the poll object insted?

murillo128 (changes requested):

src/EventLoop.cpp
```diff
@@ -179,6 +172,10 @@ bool EventLoop::Start(std::function<void(void)> loop)
 	TRACE_EVENT("eventloop", "EventLoop::Start(loop)");
 	Debug("-EventLoop::Start()\n");
 	
+	auto cleanUp = std::make_shared<ScopedCleanUp>([this](){
```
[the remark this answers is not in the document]
harryz2000: I am doing the clean up in this way to clean up whenever this function returns upon an error or when the thread loop exits if it is started. Note previously we didn't close the all pipes if one pipe is invalid. For example the code at line 193, it should be:
```
    if (pipe[0]==FD_INVALID || pipe[1]==FD_INVALID)
   {
             if (pipe[0] != FD_INVALID) close(pipe[0));
             if (pipe[1] != FD_INVALID) close(pipe[1));

            return Error("...");
   }

```
We basically need to this for every error cases. Using the ScopedCleanUp would make this kind of handling easier.

src/EventLoop.cpp
```diff
@@ -196,9 +193,17 @@ bool EventLoop::Start(std::function<void(void)> loop)
...
-	//Store socket
-	this->fd = FD_INVALID;
+	
+	
+	if (!poll->AddFd({Poll::PollFd::Category::Signaling, pipe[0]}))
+	{
+		return Error("Failed to add signaling fd to event poll\n");
+	}
+	
+	if (!poll->SetEventMask({Poll::PollFd::Category::Signaling, pipe[0]}, Poll::Event::In))
```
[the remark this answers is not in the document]
harryz2000: The code assume fd is unique if category is different. For example, For IO fd, there could be a 1. For Signaling FD, there could be a 1 as well. Only passing a fd would give the Poll enough information to indentify which PollFd it is refering. Basically for Poll class, the catagory+fd identify an entry.

src/EventLoop.cpp
harryz2000: Initially I thought of that way. I even created a SignallingPoll class. But eventually I choose to make in the EvenLoop since that would make the Poll class interface not clean. For example, I would need to add interfaces such as Signal(), CleanSignal(). That is kind of against the idea to only abstract the pure Poll interfaces rather than adding specific fd related functionality. 

Basically the idea is the Poll class only care about the generic poll infrastructure, i.e. how to implement the poll. The specific usage of the Poll should be in other classes. For our case, only the EvenLoop is using the signalling, which is part of the EventLoop functionality. So I leave the code there.

include/EventLoop.h
```diff
@@ -83,23 +87,92 @@ class EventLoop : public TimeService
...
+		
+		if (eventMask.has_value())
+		{
+			return poll->SetEventMask(pfd, *eventMask);
+		}
+		
+		return true;
+	}
+	
+	inline bool RemoveFd(int fd)
```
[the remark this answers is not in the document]
harryz2000: OK

include/EventLoop.h
```diff
@@ -83,23 +87,92 @@ class EventLoop : public TimeService
...
-	{
-		return pipe;
-	}
 
 	const std::chrono::milliseconds Now();
 	
-	virtual short GetPollEvents() const { return POLLIN | POLLERR | POLLHUP; };
+	// Functions to add/remove/iterate file descriptors
+	
+	inline bool AddFd(int fd, std::optional<uint16_t> eventMask = std::nullopt)
```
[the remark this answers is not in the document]
harryz2000: OK

include/EventLoop.h
```diff
@@ -83,23 +87,92 @@ class EventLoop : public TimeService
...
+		
+		return true;
+	}
+	
+	inline bool RemoveFd(int fd)
+	{
+		return poll->RemoveFd({Poll::PollFd::Category::IO, fd});
+	}
+	
+	inline void ForEachFd(const std::function<void(int)>& func)
```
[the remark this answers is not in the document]
harryz2000: OK

src/NetEventLoop.cpp
```diff
@@ -21,6 +21,21 @@ void NetEventLoop::ClearRawTx()
 	rawTx.reset();
 }
 
+bool NetEventLoop::SetAffinity(int cpu)
+{
+#ifdef 	SO_INCOMING_CPU
+	//If got socket
+	EventLoop::ForEachFd([cpu](auto pfd) {
+		if (pfd != FD_INVALID)
+			//Set incoming socket cpu affinity
```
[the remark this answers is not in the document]
harryz2000: Actually the pfd here is a pure fd. I should make it clearer here. As the EvenLoop virtual functions  only handle IO category of fds. The base EvenLoop handle the signalling one. So here he EvenLoop::ForEachFd only call functions on IO categories.

The Idea is the signaling category of fd is private in base EventLoop. So the subclasses doesn't need to handle its logic.

src/NetEventLoop.cpp
```diff
@@ -64,10 +79,10 @@ void NetEventLoop::Send(const uint32_t ipAddr, const uint16_t port, Packet&& pac
 	Signal();
 }
 
-short NetEventLoop::GetPollEvents() const
+std::optional<uint16_t> NetEventLoop::GetPollEventMask(int fd) const
 {
 	//If we have anything to send set to wait also for write events
-	return sending.size_approx() ? POLLIN | POLLOUT | POLLERR | POLLHUP : POLLIN | POLLERR | POLLHUP;
+	return sending.size_approx() ? (Poll::Event::In | Poll::Event::Out) : Poll::Event::In;
```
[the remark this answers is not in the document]
harryz2000: The are automatically added to the event mask in SystemPoll implementation as we need them all the time.

src/NetEventLoop.cpp
```diff
@@ -222,4 +235,9 @@ void NetEventLoop::OnPollOut(int fd)
 	items.clear();
 	//Copy elements to retry
 	std::move(retry.begin(), retry.end(), std::back_inserter(items));
-}
\ No newline at end of file
+}
+
+void NetEventLoop::OnPollError(int fd, const std::string& errorMsg)
+{
+	throw std::runtime_error("Error occurred on network fd: " + errorMsg);
```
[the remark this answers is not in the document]
harryz2000: The exception throwing here is to make the EventLoop to exit. There is a try-catch section in EventLoop::Run() that contains the code to call OnPollError().  The OnPollError is an event handling already, yes. However, this error might not be fatal. 

For example, for SRTPoll, a OnPollError on one file descriptor might just mean one SRT connection is broken. In that case, we only want to drop the connection and keep the server thread running so we wouldn't throw exception for that case.

Or maybe I misunderstood the comment?

src/EventLoop.cpp
harryz2000: We could. However I prefer to separate it as especially I am expecting the poll can return an error message in case of an error. Performance-wise, using different functions would be similar. 

Plus for different Polls, the error might be got in different ways, i.e. SRTPoll would need to call functions such as srt_getsockstate(). Splitting it would be neater to me. 

For SystemPoll, we stores the events flags after wait(). So we can easily just read the flags and check the bit, which we would need to do anyway.

src/EventLoop.cpp
harryz2000: OK

src/rtmp/rtmpclientconnection.cpp
harryz2000: For existing code, I see it called when the thread starts. I kind of kept the logic. If we want to move it to poll, it seems we would call this function multiple times unless we only add it to Poll constructor. 

Or maybe we should add it to the EvenLoop() start functions? i.e. always block the signals before starting a thread. Basically all threads would have the signals blocked.

src/EventLoop.cpp
```diff
@@ -727,3 +765,15 @@ void EventLoop::ProcessTriggers(const std::chrono::milliseconds& now)
 	TRACE_EVENT_END("eventloop");
 }
 
+
+void EventLoop::CleanUp()
+{
+	//Close pipe
+	if (pipe[0]!=FD_INVALID) close(pipe[0]);
```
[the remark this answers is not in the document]
harryz2000: Yes. We can. We still need to manually reset the FileDescriptor when doing cleanUp though.

murillo128 (changes requested):

src/SystemPoll.cpp
```diff
@@ -0,0 +1,169 @@
...
+	
+	pfds.clear();
+	tempfdsDirty = true;
+	sysfdsDirty = true;
+}
+
+
+void SystemPoll::ForEachFd(std::function<void(PollFd)> func)
+{
+	if (tempfdsDirty)
```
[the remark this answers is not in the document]
harryz2000: This temporary fds is to allow the callback to remove a fd. Otherwise, the callback cannot call RemoveFd(). i.e. the following code would cause crash:

```
for (auto fd : fds)
{
      fds.erase(fd);
}
```

include/EventLoop.h
```diff
@@ -83,23 +87,69 @@ class EventLoop : public TimeService
...
 private:
+	
+	void CleanUp();
 
-	std::thread	thread;
-	int		fd		= 0;
-	int		pipe[2]		= {FD_INVALID, FD_INVALID};
-	pollfd		ufds[2]		= {};
+	std::thread			thread;
+	std::unique_ptr<FileDescriptor>	pipeFds[2];
```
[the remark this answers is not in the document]
harryz2000: OK. I will try to remove the clean up use.

src/EventLoop.cpp
```diff
@@ -548,66 +599,94 @@ void EventLoop::Run(const std::chrono::milliseconds &duration)
 	while(running && now<=until)
 	{
 		//TRACE_EVENT("eventloop", "EventLoop::Run::Iteration");
-		ufds[0].events = GetPollEvents();
-		//Clear readed events
-		ufds[0].revents = 0;
-		ufds[1].revents = 0;
+		poll->ForEachFd([this](Poll::PollFd pfd) {
+			
+			if (pfd.category == Poll::PollFd::Category::Signaling) return;
```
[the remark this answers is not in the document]
harryz2000: The `EventLoop::ForEachFd(int fd)` only provide IO fds. However, `Poll::ForEachFd(PollFd pfd)` will provide both IO and Signalling, this is to hide the signalling part from the users.

src/EventLoop.cpp
```diff
@@ -548,66 +599,94 @@ void EventLoop::Run(const std::chrono::milliseconds &duration)
...
+			now = Now();
+			
+			//UltraDebug("<EventLoop::Run() | poll timeout:%d timers:%d tasks:%d\n",timeout,timers.size(),tasks.size_approx());
+			
+			poll->ForEachFd([this](Poll::PollFd pfd) {
+				
+				auto error = poll->GetError(pfd);
+				switch (pfd.category)
+				{
+				case Poll::PollFd::Category::Signaling:
```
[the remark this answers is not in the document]
harryz2000: As above, EventLoop::ForEachFd() is different to Poll::ForEachFd(). The argument type is different. It happens the function name is same. Potentially the first one could be renamed to EventLoop::ForEachIOFd()?

src/EventLoop.cpp
```diff
@@ -548,66 +599,94 @@ void EventLoop::Run(const std::chrono::milliseconds &duration)
...
+					assert(poll->GetEvents(pfd) != Poll::Event::Out);
+					// Always clear signal
+					ClearSignal();
+					if (error)
+					{
+						Error("-EventLoop::Run() Error occured on signaling fd: %s\n", error->c_str());
+						OnSignallingError(*error);
```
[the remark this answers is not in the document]
harryz2000: I don't think we can recover from signalling error. But for some existing code, it seems the error was just ignored. But yes, we should exit on the error.

About exiting the thread, actually I think we may want to think what should we do if the thread exits. Such as when a SRT server thread exits, it couldn't accept new SRT requests. But that would be out of current PR's scope.

src/EventLoop.cpp
```diff
@@ -548,66 +599,94 @@ void EventLoop::Run(const std::chrono::milliseconds &duration)
...
+			}
+			
+			//Update now
+			now = Now();
+			
+			//UltraDebug("<EventLoop::Run() | poll timeout:%d timers:%d tasks:%d\n",timeout,timers.size(),tasks.size_approx());
+			
+			poll->ForEachFd([this](Poll::PollFd pfd) {
+				
+				auto error = poll->GetError(pfd);
```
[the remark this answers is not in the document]
harryz2000: We could change to that way. But I personally don't like to add event masks for errors, as we always need to add errors. For SRTPoll, we even don't have a way to enable/disable the error mask as it is always reported. And we still need to report error messages. So we need still have following functions.

```
uint64_t GetEvents(fd);
std::string GetErrorMessage(fd);
```
Or:

`std::pair<uint64_t, std::string> GetEvents(fd);`

include/EventLoop.h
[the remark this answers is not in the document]
harryz2000: I changed to clear the registered fds before adding signalling fd so we can remove the clean up.
The purpose is to make the Start() can be called multiple times without duplicated fd adding.

src/EventLoop.cpp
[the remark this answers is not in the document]
harryz2000: I was trying to hide the category part for EventLoop users. But happy to change the EventLoop member functions to:
AddIOFd()
RemoveIOFd()
ForEachIOFd()

src/EventLoop.cpp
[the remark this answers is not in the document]
harryz2000: I would prefer to use the second format:
`std::pair<uint16_t, std::string> GetEvents(fd);`
Or:
`std::variant<uint16_t, std::string> GetEvents(fd);`
While I still keep only two event types. The error would be returned as an string, rather than a flag.
```
enum Event
{
  In = 0x1,
  Out = 0x2,
};
```
How do think?

harryz2000:
> that pattern is really ackwards, exception handling should only be used for exceptional error handling not flow control.

The exception means we need to exit the thread. To me, it can be thought of an exceptional case. Normally we don't want to exit the thread. 

There are potential other ways:
1. Return different value, such as bool. such as `bool OnPollError()`, this is sometimes confusing.
2. Passing a reference to capture the intended behaviour from callbacks, such as void OnPollError(bool& exit);

The options might not be as clean as the exception.

harryz2000:
> > that pattern is really ackwards, exception handling should only be used for exceptional error handling not flow control.
> 
> The exception means we need to exit the thread. To me, it can be thought of an exceptional case. Normally we don't want to exit the thread.
> 
> There are potential other ways:
> 
> 1. Return different value, such as bool. such as `bool OnPollError()`, this is sometimes confusing.
> 2. Passing a reference to capture the intended behaviour from callbacks, such as void OnPollError(bool& exit);
> 
> The options might not be as clean as the exception.

I would prefer to use option 2 if not using exception. Any suggestions?

src/EventLoop.cpp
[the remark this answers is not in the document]
murillo128: but it would be notified on the loop exit callback, right?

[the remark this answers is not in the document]
harryz2000: Yeah. We could use the loop exit callback. But currently that callback doesn't contain the error code unless we store the error code somewhere, which we could do. I thought of that. It is actually possibly cleaner. However I feel that change would be kind of out of current scope. Not sure we should do the change under current PR.

include/Poll.h
```diff
@@ -0,0 +1,131 @@
...
+ */
+template<>
+struct std::hash<Poll::PollFd>
+{
+	size_t operator()(const Poll::PollFd& fd) const
+	{
+		// While this hash function works when size of int is not 4, but it might be better way to
+		// calculate in that case. Assert here for reminder.
+		static_assert(sizeof(int) == 4);
+		return (uint64_t(static_cast<std::underlying_type<Poll::PollFd::Category>::type>(fd.category)) << 32) + uint32_t(fd.fd);
```
yezhan10, line 126: maybe consider to use std::underlying_type_t for better readability?

src/EventLoop.cpp
```diff
@@ -192,37 +188,67 @@ bool EventLoop::Start(std::function<void(void)> loop)
...
+
 	//Check values
 	if (pipe[0]==FD_INVALID || pipe[1]==FD_INVALID)
 		//Error
 		return Error("-EventLoop::Start() | could not start pipe [errno:%d]\n",errno);
-			
-	//Store socket
-	this->fd = FD_INVALID;
+	
+	// Remove all signaling fds
```
yezhan10, line 200: why the comment says remove all signaling fds? should this ForEachFd() remove both IO and Signaling fds?

include/Poll.h
yezhan10: Or maybe worth of creating something like this?: 
`template<typename E>
constexpr auto
	toUType(E enumerator) noexcept
	{
		return static_cast<std::underlying_type_t<E>>(enumerator);
	}`

harryz2000: Would change to std::underlying_type_t. 
I probably wouldn't add the help function as currently it is just used in very few places.

include/Poll.h
```diff
@@ -0,0 +1,131 @@
...
+ * GetEvents()/GetError().
+ */
+class Poll
+{
+public:
+	
+	/**
+	 * Event types.
+	 */
+	enum Event : uint16_t
```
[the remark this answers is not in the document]
harryz2000: As discussed, it is used as constants. I would prefer to use the unscoped one here for easy to calculate.

src/EventLoop.cpp
harryz2000: Good catch! Indeed I meant to remove signaling fds. Other fds would be managed by user.

src/EventLoop.cpp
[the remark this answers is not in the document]
murillo128: we have added those callbacks in this PR, so we can do wathever we want

src/SystemPoll.cpp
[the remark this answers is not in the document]
murillo128: I don't think we should allow the callback to modify the fds

src/EventLoop.cpp
```diff
@@ -192,37 +188,70 @@ bool EventLoop::Start(std::function<void(void)> loop)
 #else
 	pipe[0] = pipe[1] = eventfd(0, EFD_NONBLOCK);
 #endif	
+
+	pipeFds[0] = FileDescriptor(pipe[0]);
```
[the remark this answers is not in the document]
harryz2000: The move is not necessary as the FileDescriptor is already an RValue. I think assigning a int to the FileDescriptor wouldn't work as it doesn't have a operator=(int) override. But I will try.

src/EventLoop.cpp
```diff
@@ -192,37 +188,70 @@ bool EventLoop::Start(std::function<void(void)> loop)
...
 		return Error("-EventLoop::Start() | could not start pipe [errno:%d]\n",errno);
-			
-	//Store socket
-	this->fd = FD_INVALID;
+	
+	// Remove all signaling fds
+	poll->ForEachFd([this](Poll::PollFd pfd){
+		if (pfd.category == Poll::PollFd::Category::Signaling)
+		{
+			(void)poll->RemoveFd(pfd);
```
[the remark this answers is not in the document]
harryz2000: Because I don't want to clean all fds. It is not illegal to add fd first and then call the Start() function. For example,
```
loop->AddFd(fd);
loop->Start();
loop->Stop();
loop->Start();
```
I don't want the Start() to clear the added fd by caller. It only clears the signaling one it added.

harryz2000:
> I don't think we should allow the callback to modify the fds
But we do need to remove for SRTPoll. Such as when it receive an OnPollError() it means the SRT socket was closed. We need to remove it from the map. Any idea to do that?

include/ScopedCleanUp.h
```diff
@@ -0,0 +1,27 @@
+#ifndef SCOPEDCLEANUP_H
```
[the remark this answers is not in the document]
harryz2000: yep

src/SystemPoll.cpp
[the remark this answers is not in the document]
harryz2000: But we do need to remove for SRTPoll. Such as when it receive an OnPollError() it means the SRT socket was closed. We need to remove it from the map. Any idea to do that?

src/EventLoop.cpp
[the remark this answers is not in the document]
murillo128: are you sure that we are not going to `dup` the `fd` if not moving?

src/EventLoop.cpp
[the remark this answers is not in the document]
murillo128:
```suggestion
			poll->Clear(Poll::PollFd::Category::Signaling);
```

src/EventLoop.cpp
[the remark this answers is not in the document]
murillo128: add a filtering criteria for `ForEachFd` then better

src/EventLoop.cpp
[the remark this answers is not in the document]
murillo128: we should not report errors as strings

murillo128 (changes requested):

src/EventLoop.cpp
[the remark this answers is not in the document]
harryz2000: Take SRT for example, I am not sure a SRTSocket value would be possibly same as an signalling fd value. That is why a category is used to tell which type it is.

src/EventLoop.cpp
```diff
@@ -192,37 +188,70 @@ bool EventLoop::Start(std::function<void(void)> loop)
...
 	//Check values
 	if (pipe[0]==FD_INVALID || pipe[1]==FD_INVALID)
 		//Error
 		return Error("-EventLoop::Start() | could not start pipe [errno:%d]\n",errno);
-			
-	//Store socket
-	this->fd = FD_INVALID;
+	
+	// Remove all signaling fds
+	poll->ForEachFd([this](Poll::PollFd pfd){
```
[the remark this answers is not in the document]
harryz2000: Would change to this way.

murillo128:
one thing I was thinking,  is there any reason why we have to injection the signaling pipe from the evenloop and not handle it inside the Poll directly?

the only thing that the evenloop needs is a way to mask the poll wait to exit.

harryz2000:
> one thing I was thinking, is there any reason why we have to injection the signaling pipe from the evenloop and not handle it inside the Poll directly?
> 
> the only thing that the evenloop needs is a way to mask the poll wait to exit.

Good thought. Yes. We could optimise this even further. One question is if there is another way other than using a signalling fd for break the poll. I didn't investigate too much about that yet as I wanted to limit the scope of the change. If there is no other way, we could create a class such as SignallingPoll<SystemPoll> to wrap the signalling part. But that is basically a duplication of the logic in EventLoop. I still think the basic Poll interface shouldn't have a Signal/ClearSignal function because it is the most basic functionality. For example the SignalingPoll could be a template class to add the new functions:

```
template<typename PollClass>
class SignallingPoll : public PollClass
{
public:
       void Signal();
       void ClearSignal();
};
```

murillo128:
my point was about moving the eventfd inside the poll object,  it is only used fircsignaling the poll to exit the wait function,  so no sure why it needs to be setup at the event loop

harryz2000:
> my point was about moving the eventfd inside the poll object, it is only used fircsignaling the poll to exit the wait function, so no sure why it needs to be setup at the event loop

The main rationale is the Poll interface is to abstract the basic poll functions for system poll and SRT poll. Both the system poll and SRT poll functions don't have the functions to break the wait, as far as I could see for now.

If we want to break the wait, i.e. by adding signalling, that is something functionality we are adding to the basic functionality. So as per the rationale I put the logic into another class as the functionality is an add-on. This class can be a separate class such as SignallingPoll. It is just currently it is in EventLoop.

Does it makes sense?

murillo128:
I don't think we need so many abstraction layers providing no actual benefit.  We could have some helpers to avoid duplicating code, but I don't see the value of the SignallingPoll

harryz2000:
> I don't think we need so many abstraction layers providing no actual benefit. We could have some helpers to avoid duplicating code, but I don't see the value of the SignallingPoll

OK. I can move the code to the Poll base class then.

src/EventLoop.cpp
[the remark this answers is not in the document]
harryz2000: Confirmed. As above mentioned, it would only trigger the move assign, wouldn't call dup.

src/EventLoop.cpp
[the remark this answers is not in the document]
harryz2000: Added a Setup function in Poll() to only add signaling fd if not setup.

include/Poll.h
harryz2000: added the helper function. After changing the way to set exit loop code. Conversion from an scoped enum becomes more.

src/EventLoop.cpp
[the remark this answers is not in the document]
harryz2000: done

src/EventLoop.cpp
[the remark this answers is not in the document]
harryz2000: I added a SetStopping(exitCode) in EventLoop to flag the thread needs to be stopped.

src/EventLoop.cpp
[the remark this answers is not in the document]
harryz2000: Change the function to :
`std::pair<uint16_t, int> GetEvents(fd);`
The first of the returned pair is events flags, the second is error code.

src/EventLoop.cpp
```diff
@@ -192,37 +188,70 @@ bool EventLoop::Start(std::function<void(void)> loop)
...
+		if (pfd.category == Poll::PollFd::Category::Signaling)
+		{
+			(void)poll->RemoveFd(pfd);
+		}
+	});
+	
+	if (!poll->AddFd({Poll::PollFd::Category::Signaling, pipe[0]}))
```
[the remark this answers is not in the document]
harryz2000: Done

src/Poll.cpp
```diff
@@ -0,0 +1,85 @@
+#include "Poll.h"
+#include "log.h"
+
+#include <sys/eventfd.h>
+#include <thread>
+
+bool Poll::Setup()
```
murillo128, line 7: why can't this happen on constructor?

include/Poll.h
```diff
@@ -0,0 +1,147 @@
...
+
+	/**
+	 * Setup poll
+	 */
+	bool Setup();
+
+	/**
+	 * Signaling functions.
+	 */
+	void Signal();
```
murillo128, line 69: I would prefer to keep this functions pure virtual and move the pipe to the SystemPoll instead

include/Poll.h
```diff
@@ -0,0 +1,147 @@
...
+	void ClearSignal();
+	
+	/**
+	 * Add a file descriptor
+	 * 
+	 * @return Whether the file descriptor was added successfully
+	 */
+	virtual bool AddFd(PollFd::Category category, int fd) = 0;
```
murillo128, line 77: as we are moving the pipe to the SystemPoll, there is no need to have the `PollFd::Category` anymore

murillo128 (changes requested):

src/Poll.cpp
harryz2000: because it requires to call virtual functions, which cannot be called in constructor.

include/Poll.h
harryz2000: So we duplicate the code in SRTPoll? The SRTPoll would use same logic.

include/Poll.h
harryz2000: It seems you would like to implement SystemPoll and SRTPoll separately by duplicating the signalling logic?

src/Poll.cpp
murillo128: I mean, if we move the pipe to the SystemPoll, we can move this to its constructor.

include/Poll.h
murillo128: yes, duplication would be minimal. You can create a SignalingPipe helper class to wrap the eventfd/pipe creation

include/Poll.h
murillo128: SRTPoll is already a separate class, what I want to avoid is to have to deal with the signaling at the EventLoop level

include/Poll.h
harryz2000: OK. It is not my personal preference but happy to do that.

include/Poll.h
harryz2000: OK

src/EventLoop.cpp
```diff
@@ -548,66 +512,75 @@ void EventLoop::Run(const std::chrono::milliseconds &duration)
...
-			//Exit
-			break;
-		}
 		
-		//Read first
-		if (ufds[0].revents & POLLIN)
-		{
-			OnPollIn(ufds[0].fd);
-		}
+		auto [events, error] = poll->GetSignallingEvents();
```
murillo128, line 541: could we return the error better directly on the `Wait()` ?

src/EventLoop.cpp
```diff
@@ -548,66 +512,75 @@ void EventLoop::Run(const std::chrono::milliseconds &duration)
...
-		//Check read is possible
-		if (ufds[0].revents & POLLOUT)
+		if (events != 0 || error != 0)
+			poll->ClearSignal();
+			
+		if (error != 0)
 		{
-			OnPollOut(ufds[0].fd);
+			Error("-EventLoop::Run() Error occured on signaling fd: %d\n", error);
+			SetStopping(ToUType(PredefinedExitCode::SignalingError));
```
murillo128, line 549: shouldn't we exit on error?

src/EventLoop.cpp
murillo128: we don't need the events from the signaling pipe at all, right?

src/EventLoop.cpp
```diff
@@ -548,66 +512,75 @@ void EventLoop::Run(const std::chrono::milliseconds &duration)
...
-		if (ufds[0].revents & POLLIN)
-		{
-			OnPollIn(ufds[0].fd);
-		}
+		auto [events, error] = poll->GetSignallingEvents();
 		
-		//Check read is possible
-		if (ufds[0].revents & POLLOUT)
+		if (events != 0 || error != 0)
+			poll->ClearSignal();
```
murillo128, line 544: shouldn't we better clear the signal last as it was done on the original code?

src/SystemPoll.cpp
```diff
@@ -0,0 +1,190 @@
...
+}
+
+std::pair<uint16_t, int> SystemPoll::GetSignallingEvents() const
+{
+	return GetEvents(signalling.GetFd());
+}
+
+bool SystemPoll::Wait(uint32_t timeOutMs)
+{
+	if (sysfdsDirty)
```
murillo128, line 157: as we are not iterating over the syspfds, there is no reason to have a separate vector for that, right?

murillo128 (changes requested):

src/EventLoop.cpp
harryz2000: No. I can remove the events

src/EventLoop.cpp
harryz2000: Yes. It would exit. The SetStopping would cause the thread exit. Or could you clarify the comment?

src/EventLoop.cpp
harryz2000: It seems to me it wouldn't impact too much as we would exit anyway. I will move it later along changing the GetSignalEvents() return value.

src/SystemPoll.cpp
harryz2000: The syspfds is used for poll() function to retrieve events. As for managing the fds, we use a map rather than an array. So we have to convert it to an array (vector) before calling the poll() function. The flag sysfdsDirty is to ensure the vector is only recreated when something changed.  I couldn't we could remove it. Or I missed anything?

src/EventLoop.cpp
harryz2000: I further removed the GetSignallingEvents(), and added the logic to the wait function. 
Also removed the ClearSignal function, now the clear is automatically done in the wait function.

src/EventLoop.cpp
harryz2000: Moved the clear to Wait() function so it would be automatically cleared. I don't see the clear has to be after the Error logging and set stopping function. Making the ClearSignal() in wait function make the interface simpler to me.

src/Poll.cpp
harryz2000: Done

src/PollSignalling.cpp
```diff
@@ -0,0 +1,74 @@
...
+	if (signaled.test_and_set() || !pipeFds[1].isValid())
+		//No need to do anything
+		return;
+			
+	//We have signaled it above
+	//worst case scenario is that race happens between this to points
+	//and that we signal it twice
+	
+	//Write to tbe pipe, and assign to one to avoid warning in compile time
+	one = write(pipeFds[1],(uint8_t*)&one,sizeof(one));
```
bcostdolby, line 56: Just curious why we need to write a 64bit int and not just 1 byte?

bcostdolby, line 56: Should we also check for errors on return from write and throw or at least log an error? In theory they probably wont happen but if they do would be very useful to know.

src/PollSignalling.cpp
```diff
@@ -0,0 +1,74 @@
...
+}
+
+void PollSignalling::ClearSignal()
+{
+	if (pipeFds[0].isValid())
+	{
+		uint64_t data = 0;
+		//Remove pending data on signal pipe
+		while (read(pipeFds[0], &data, sizeof(uint64_t)) > 0)
```
bcostdolby, line 65: Not all errors are the same. I think we should retry on EINTR at least. Possibly log error on other unexpected errors

src/SystemPoll.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+#include <errno.h>
+#include <unistd.h>
+#include <fcntl.h>
+
+SystemPoll::SystemPoll()
+{
+	if (!SystemPoll::AddFd(signalling.GetFd()))
+	{
+		throw std::runtime_error("Failed to add signaling fd to event poll\n");
```
bcostdolby, line 16: Worth adding the strerror reason as well as that is more important than the op failed. Why did it fail

src/SystemPoll.cpp
```diff
@@ -0,0 +1,206 @@
...
+	}
+	
+	if (!SystemPoll::SetEventMask(signalling.GetFd(), Poll::Event::In))
+	{
+		throw std::runtime_error("Failed to set event mask\n");
```
bcostdolby, line 21: Worth adding the strerror reason as well as that is more important than the op failed. Why did it fail

src/SystemPoll.cpp
```diff
@@ -0,0 +1,206 @@
...
+	
+	// Copy back
+	for (size_t i = 0; i < syspfds.size(); i++)
+	{
+		if (syspfds[i].fd == signalling.GetFd())
+		{
+			auto revents = syspfds[i].revents;
+			if ((revents & POLLHUP) || (revents & POLLERR))
+			{
+				signalling.ClearSignal();
```
bcostdolby, line 191: Should we still iterate over all the other items and return at the end?

Some poll events are edge triggered instead of level triggered (not sure we make use of this though) and if we dont handle them we lose valuable information. 

I dont think that we are using any edge triggered events but if we do then this will be broken

include/PollSignalling.h
```diff
@@ -0,0 +1,28 @@
...
+	void Signal();
+	void ClearSignal();
+	
+	inline int GetFd() const
+	{
+		return pipeFds[0];
+	}
+	
+private:
+	FileDescriptor pipeFds[2];
```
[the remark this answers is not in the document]
harryz2000: Not sure that would be better. I would prefer to leave it as it is.

src/PollSignalling.cpp
harryz2000: Not sure. Maybe @murillo128 knows the reason. The code was copied from existing code. Again, I would prefer to limit the scope of this PR.

harryz2000: As above, the code was copied from existing code. Again, I would prefer to limit the scope of this PR. I would make notes for the comments.

src/PollSignalling.cpp
harryz2000: As above, the code was copied from existing code. Again, I would prefer to limit the scope of this PR. I would make notes for the comments.

src/SystemPoll.cpp
harryz2000: I don't see much benefits since we would exit the thread anyway. Event the events was set, the return value is still indicating there is an error. The revents wouldn't be regarded as valid. On the other hand, this error shouldn't happen unless OS got an error.

I don't think we would use edge triggered. As per https://stackoverflow.com/questions/15072165/is-poll-an-edge-triggered-function, it seems the poll function is always level triggered. Epoll might work in different modes. But we are not using Epoll.

include/tools.h
```diff
@@ -533,4 +534,29 @@ inline void SyncWriteUint128(__uint128_t *dst, __uint128_t value)
...
+ * 
+ * @param fmt Format with specificers
+ * @param ... Aguments
+ * @return The result string
+ */
+inline std::string FormatString(const char* fmt, ...)
+{
+	va_list ap;
+	va_start(ap, fmt);
+	auto sz = std::vsnprintf(nullptr, 0, fmt, ap);
```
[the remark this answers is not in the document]
harryz2000: It seems current implementation has no problem. I added the va_start/va_end as it looks that is the right way in theory, yes.

src/SystemPoll.cpp
harryz2000: Adding more info is always better, yes. I added logs in the function to show the details rather than put it in this exception. I think that would make the interface simpler. Otherwise a error code might not be easy to reflect different situations from different Poll implementations. It would up to the concrete class to log the error details.

src/SystemPoll.cpp
harryz2000: Adding more info is always better, yes. I added logs in the function to show the details rather than put it in this exception. I think that would make the interface simpler. Otherwise a error code might not be easy to reflect different situations from different Poll implementations. It would up to the concrete class to log the error details.

src/SystemPoll.cpp
bcostdolby: Ahh yes. My recollection was from before using epoll.

src/PollSignalling.cpp
murillo128: eventfd used 8bytes reads/writes

include/FileDescriptor.h
```diff
@@ -39,20 +39,20 @@ class FileDescriptor {
 	{
 		fd = dup(other.fd);
 	}
-	FileDescriptor& operator=(FileDescriptor other)
+	FileDescriptor& operator=(const FileDescriptor& other)
 	{
-		swap(*this, other);
+		fd = dup(other.fd);
```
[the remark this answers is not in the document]
mildsunrise: I'm looking at this and don't understand
1. what's wrong with the previous code? it's the [copy and swap idiom](https://stackoverflow.com/a/3279550/710951) and there's usually little reason to change it
2. doesn't the new code fail to close the previous FD, therefore leaking it?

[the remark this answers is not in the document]
mildsunrise: regarding error handling of dup(), the only errors it can raise are EBADF (if the passed descriptor is not open) and EMFILE (if we've reached the limit for open file descriptors in the process). EBADF should never happen unless there's a bug, and if EMFILE happens I'm pretty sure the media server would just collapse entirely so no point in doing anything smart, I'm okay with just asserting and crashing otherwise

src/PollSignalling.cpp
bcostdolby: yeah I read the docs for the event in more detail after writing the comment and forgot to resolve it.

include/FileDescriptor.h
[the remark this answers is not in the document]
harryz2000: The reason is the previous code:
`FileDescriptor& operator=(FileDescriptor other)`
was not a entirely a copy operator.
It causes compile errors in some cases, i.e. when other is a R value iirc, as the compiler wouldn't know which one it should choose between it and 
`FileDescriptor& operator=(FileDescriptor&& other)`
While as a copy assignment, it should not accept a R value. And it functionality should be same as copy constructor. 
`FileDescriptor(const FileDescriptor& other)`
Otherwise, it would surprise the users.
https://en.wikipedia.org/wiki/Rule_of_three_(C%2B%2B_programming)#Rule_of_five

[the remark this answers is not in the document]
mildsunrise: thanks, that explains it. shouldn't we just remove `FileDescriptor& operator=(FileDescriptor&& other)` then?
as the post I linked explains, `FileDescriptor& operator=(FileDescriptor other)` acts as both a copy operator and a move operator, because the compiler will move construct into `FileDescriptor other` if it's an rvalue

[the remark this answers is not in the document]
harryz2000: That might resolve the compiling issue, yes. However I think it is unusual to do that way and it may not be good for performance. For rule of five, all the five functions should be declared explicitly. Using `FileDescriptor& operator=(FileDescriptor other)` for both L and R values would introduce unnecessary constructing:

For example, when it is assigned an R value, it would call:
```
FileDescriptor(FileDescriptor&&)
FileDescriptor& operator=()
```
when it is assigned to a L value, it would call:
```
FileDescriptor(const FileDescriptor&)
FileDescriptor& operator=()
```
If we declare` operator=` for L and R value separately, the constructor wouldn't be called.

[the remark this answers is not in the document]
mildsunrise: the copy-and-swap idiom [is](https://www.geeksforgeeks.org/copy-swap-idiom-c/) [fairly](https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Copy-and-swap) [popular](https://www.reddit.com/r/cpp/comments/1am9jzm/copy_and_swap_idiom/). it is *much* less error prone (especially with pitfalls of assigning something to itself, or forgetting to free current resources like what happened here), less code to write, and provides strong exception safety.

> it may not be good for performance

don't underestimate compilers! conceptually you may be calling those 2 things (the constructor and the assign operator) but the compiler knows the definition of both and can [elide everything away, so no intermediary object is actually created in memory](https://godbolt.org/z/P3hYTsWcT).

for objects that can re-use its existing resources (like allocations) to host the copy, then copy-and-swap can be wasteful, but that doesn't apply here.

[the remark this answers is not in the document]
harryz2000: Thanks for providing the examples. Yes. It is not a surprise that the compiler could optimise the code to very deep level. The idiom looks elegant and no performance impact in some cases. My concern is the benefit, especially the performance impact, might not apply to all the classes. imho, it would be better to have consistent code style so it could be easily reviewed and the behaviour would be more predictable. For this specific case, I agree the copy-and-swap would work.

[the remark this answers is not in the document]
mildsunrise: we can have a vote on this, but given how tricky it is to implement everything in the rule of five correctly, my stance is to always do copy-and-swap except on those specific cases where there is a performance impact and it is high enough... because every single time I've reviewed a RAII class that was not using copy-and-swap, the assignment operator was not implemented correctly

PR #351, opened 2024-09-26
add unit tests for h264 sps/pps parsing

amaha877:
Adding some tests for h264 sps pps parsing


amaha877:
I've encapsulated the sps/pps data in structs so that its easy for the tests to validate (instead of relying on stdout `Dump()` )

src/h264/h264.h
```diff
@@ -12,11 +12,6 @@ class H264SeqParameterSet
 		RbspReader reader(buffer, bufferSize);
 		RbspBitReader r(reader);
 
-		for (int i = 0; i < bufferSize; i++)
-		{
-			printf("0x%x, ", buffer[i]);
```
murillo128: you can use 

::Dump(buffer, bufferSize)

instead

src/h264/h264.h
```diff
@@ -110,76 +110,84 @@ class H264SeqParameterSet
...
+		bool			mb_adaptive_frame_field_flag = false;
+		bool			direct_8x8_inference_flag = false;
+		bool			frame_cropping_flag = false;
+		DWORD			frame_crop_left_offset = 0;
+		DWORD			frame_crop_right_offset = 0;
+		DWORD			frame_crop_top_offset = 0;
+		DWORD			frame_crop_bottom_offset = 0;
+		//bool			vui_parameters_present_flag = false;
+		bool			separate_colour_plane_flag = 0;
+	}data;
```
murillo128, line 146: moving all the attributes to an internal field is not in scope of this task, please keep changes to the minimum for easier review and not disruption othe codebasr

murillo128 (changes requested):

murillo128:
better either make the attributes public or try adding the test class as friend (not sure if possible)

src/h264/h264.h
amaha877: Okay, I'll make it public.

src/h264/h264.h
```diff
@@ -110,7 +110,39 @@ class H264SeqParameterSet
 		//OK
 		return true;
 	}
+
 public:
+	BYTE			profile_idc = 0;
```
murillo128, line 115: could you move back this code block to its previous position? it is harder to review if I have to go line by mine to verify if there has been any unintended change committed

murillo128 (approved):
lgtm

amaha877:
Added the fixes. @murillo128 please take a look.

include/bitstream/BitReader.h
```diff
@@ -168,7 +168,7 @@ class BitReaderBase
 		//Get code num
 		DWORD codeNum = GetExpGolomb();
 		//Conver to signed
-		return codeNum & 0x01 ? codeNum >> 1 : -(codeNum >> 1);
+		return codeNum & 0x01 ? (codeNum+1) >> 1 : -(codeNum >> 1);
```
amaha877, line 171: Verified this with libav as well as h264NAL tool

src/h264/H26xNal.h
```diff
@@ -72,6 +72,9 @@ class RbspReader
 
 	inline void Skip(size_t num)
 	{
+		//Don't update pos if we have not consumed any bytes
+		if (!num) return;
```
amaha877, line 76: if we don't do this check, we keep incrementing the position of the read pointer and eventually overflow the entire buffer and thereby causing an exception. So if we haven't consumed any whole bytes, then nothing to Skip.

PR #353, opened 2024-09-27
MPEGTS encoding and Packetizer class. Prepare for SRT Client implementation.

harryz2000:


harryz2000:
I am planning to add more unit tests.

src/EventLoop.cpp
```diff
@@ -569,7 +569,7 @@ void EventLoop::Signal()
 	uint64_t one = 1;
 	
 	//If we are in the same thread or already signaled and pipe is ok
-	if (std::this_thread::get_id()==thread.get_id() || signaled.test_and_set() || *pipeFds[1] == FD_INVALID)
+	if (std::this_thread::get_id()==thread.get_id() || signaled.test_and_set() || !pipeFds[1])
```
murillo128, line 572: this shouldn't be here

src/FrameDelayCalculator.cpp
```diff
@@ -8,23 +8,6 @@ using namespace std::chrono;
...
 
-
-/**
- * Convert timestamp from one clock rate to another
- * 
- * @param ts The input timestamp
- * @param originalRate The clock rate of the input timestamp
- * @param targetRate The target clock rate
- * 
- * @return The timestamp basing on the target clock rate
```
murillo128, line 19: this change does not belong here

src/mpegts/psi.cpp
```diff
@@ -95,30 +204,29 @@ Table Table::Parse(BufferReader& reader)
 	return table;
 }
 
-std::vector<Table> ParsePayloadUnit(BufferReader& reader)
```
murillo128, line 98: why has this been removed?

src/mpegts/psi.cpp
```diff
@@ -140,35 +248,35 @@ ProgramAssociation ProgramAssociation::Parse(BufferReader& reader)
 	return programAssociation;
 }
 
-std::vector<ProgramAssociation> ProgramAssociation::ParsePayloadUnit(BufferReader& reader)
```
murillo128, line 143: why has this been removed?

include/Packetizer.h
```diff
@@ -0,0 +1,174 @@
+#ifndef PACKETIZER_H
```
murillo128, line 1: this seems too mpegts specific

include/Encodable.h
```diff
@@ -0,0 +1,25 @@
...
+
+class Encodable
+{
+public:
+	virtual ~Encodable() = default;
+	
+	/**
+	 * Encode the content
+	 */
+	virtual void Encode(BufferWritter& writer) const = 0;
```
murillo128, line 16: We have always been called this method `Serialize`

include/Encodable.h
```diff
@@ -0,0 +1,25 @@
...
+	
+	/**
+	 * The memory size needed for encoding. It must match
+	 * the exact buffer used after encoding the content.
+	 */
+	virtual size_t Size() const = 0;
```
murillo128, line 22: Also, we have called this method `GetSize()`

include/Encodable.h
```diff
@@ -0,0 +1,25 @@
+#ifndef ENCODABLE_H
+#define ENCODABLE_H
+
+#include <cstddef>
+
+class BufferWritter;
+
+class Encodable
```
murillo128, line 8: `Serializable` seems a better name

murillo128 (changes requested):

src/mpegts/psi.cpp
harryz2000: The code is not used at the moment and I think it was causing a issue in my development.  I would add it back as it seems no problem any more.

include/Encodable.h
harryz2000: OK. Will rename

include/Encodable.h
harryz2000: OK

include/Encodable.h
harryz2000: OK

include/Packetizer.h
harryz2000: I tried to make it generic. Currently it works for mpegts specifically. However, it could be used for any Messages that implements the Encodable/Serializable interface. We could see if we need to make it more generic when we found other use cases

src/EventLoop.cpp
harryz2000: Could you clarify this comment? thx

src/FrameDelayCalculator.cpp
harryz2000: It seems this is a change for another PR. will fix

src/EventLoop.cpp
murillo128: I mean that this change is not related to mpegts encoding, probably a bad rebase or similar

include/Packetizer.h
murillo128: I mean, that I don't see it usable for anything different than mpegts encoding, probably would need to wait until the SRT PR is ready for review to see how it is used

src/EventLoop.cpp
harryz2000: That is right. It should belong to another PR :)

include/Packetizer.h
```diff
@@ -0,0 +1,174 @@
...
+public:
+	
+	/**
+	 * Constructor
+	 * 
+	 * @param maxMessageQueueSize The max number of messages the packetizer can keep before they are packetized.
+	 */
+	Packetizer(size_t maxMessageQueueSize) : messages(maxMessageQueueSize, false), buffer(0)
+	{
+		static_assert(std::is_base_of_v<Serializable, T>);
```
yezhan10, line 49: do you think something like this is better?
`template<typename T, typename = std::enable_if_t<std::is_base_of_v<Serializable, std::decay_t<T>>>>`
This will force the T to be derived from Serializable as well? But i am not sure which error message is better in case of misuse.

harryz2000: Yes. That should work as well and probably better in some way. However for code readability, I feel the static_assert is even easier to read. Ideally, we should use concepts for that. But we have to upgrade to C++ 20 for it.
```
template <typename T> requires std::derived_from<T, Serializable>
```
For now, I would prefer to leave it as it is.

yezhan10: did not know that c++20 would do something like that, thanks for sharing the info!

include/Packetizer.h
harryz2000: Yeah. I will soon make the srt PR available. You can review this code after that. I would prefer keep this code in current PR to avoid frequent branch switching/merging.

src/EventLoop.cpp
harryz2000: fixed

src/mpegts/psi.cpp
```diff
@@ -37,17 +90,66 @@ SyntaxData SyntaxData::Parse(BufferReader& reader)
 
 	size_t dataSize = reader.GetLeft() - 4;
 	syntaxData.data			= BufferReader(reader.GetData(dataSize), dataSize);
+	
+	return syntaxData;
+}
 
-	syntaxData.crc32		= reader.Get4();
```
murillo128, line 41: what has happened to the `crc32`?

src/mpegts/psi.cpp
murillo128: please, don't resolve the conversation until you add this code back

harryz2000: Ah. I added back the` std::vector<Table> ParsePayloadUnit(BufferReader& reader);` one. I thought it was this one. Sorry. Will add this one back as well.

src/mpegts/psi.cpp
harryz2000: It was moved to the Table pasing. The CRC32 is not only for the syntax data part.
https://github.com/medooze/media-server/blob/beebaddd5964c6466c687e771d7ada01001202fc/src/mpegts/psi.cpp#L192
This is to make SyntaxData encoding is consistent with parsing. See:
https://github.com/medooze/media-server/blob/beebaddd5964c6466c687e771d7ada01001202fc/src/mpegts/psi.cpp#L127

src/mpegts/psi.cpp
harryz2000: added back

include/Packetizer.h
```diff
@@ -0,0 +1,174 @@
...
+ * 
+ * One generated packet could contain bytes from mutliple messages. The behaviour can be set by the
+ * forceSeparatePacket flag. When it is true, the begining of current message will be packetized in
+ * next packet. Otherwise, the bytes would be appended to current packet if current packet is partly
+ * filled.
+ * 
+ * The Packetizer uses a cicular queue to store added messages. The limit of the queue size is set during
+ * construction. If the queue is full, the earlest message would be dropped.
+ * 
+ * @tparam T The message type. It must be derived from Serializable.
```
murillo128, line 35: if it must be derived from `Serializable` why do we need it as template variable?

include/Packetizer.h
```diff
@@ -0,0 +1,174 @@
...
+	}
+	
+	/**
+	 * Get the next packet and write into the BufferWriter. The Packetizer will try to 
+	 * fill all the availabe space in the writer unless the left bytes of the message is
+	 * less.
+	 * 
+	 * @return size_t The size of written bytes.
+	 */
+	virtual size_t GetNextPacket(BufferWritter& writer)
```
murillo128, line 109: this method does not Get anything, rename to `SerializeNextPacket` or `WriteNextPacket` or similar

include/Packetizer.h
murillo128: after reviewing the srt PR, I don't feel that this class has any meaning outside of it, I am not even sure if it should not be just integrated inside the outgoing stream code directly.

harryz2000: It is now only used for SRT code. The class is quite generic. We could have other use. For example, I am fine if you think it is good to move it to SRT code. However imho it would be good to not include it inside the outgoing stream code. As at least this is easy for testing.

include/Packetizer.h
harryz2000: Initial implementation doesn't have this requirement. But yes, it could be changed to use Serializable instead.

include/Packetizer.h
harryz2000: OK

include/Packetizer.h
murillo128: me can keep the class in the media-server-src but moving it to the mpegts dir and renaming to something a bit more verbose, we already have RTPPacketizers, so could be easy to confuse them

include/Serializable.h
```diff
@@ -0,0 +1,25 @@
+#ifndef SERIALIZABLE_H
+#define SERIALIZABLE_H
+
+#include <cstddef>
+
+class BufferWritter;
+
+class Serializable
```
murillo128, line 8: shouldn't only the top level messages (i.e. PESMessage, PSIMessage) derive from `Serializable`  calling `virtual` methods hurts performance and prevents inlining

harryz2000: Can you clarify a bit? Do you mean make all the mpegts fields, such as mpgets::Header, not derived from Serializable? Or the PESMessage child classes, such as H26xPESMessage, not derived from Serializable? PESMessage is in another PR. But not sure what do you mean top level.

For the fields, it doesn't need to inherit from Serializable, as there is no need for inheritance there. I can remove it. For the message, I wouldn't think the inheritance would impact performance a lot as the function is only called per message, which means a frame. We could use templates instead of inheritance to get rid of the virtual functions. Do you think we could go that way or any other thoughts?

harryz2000: Removed the Serialzable inheritance from fields now.

src/mpegts/Packetizer.h
```diff
@@ -0,0 +1,171 @@
...
+ * forceSeparatePacket flag. When it is true, the begining of current message will be packetized in
+ * next packet. Otherwise, the bytes would be appended to current packet if current packet is partly
+ * filled.
+ * 
+ * The MessagePacketizer uses a cicular queue to store added messages. The limit of the queue size is set during
+ * construction. If the queue is full, the earlest message would be dropped.
+ * 
+ */
+template<typename T>
+class MessagePacketizer
```
murillo128, line 36: rename filename to match class name please

src/FrameDelayCalculator.cpp
murillo128: does this needs a rebase?

src/mpegts/mpegtscrc32.cpp
```diff
@@ -0,0 +1,62 @@
+#include "mpegtscrc32.h"
+
+namespace mpegts
+{
+
+static const uint32_t CrcTable[256] = {
```
murillo128, line 6: is this crc32 different than the one already available  https://github.com/medooze/media-server/blob/c611ec586d83527ebcba21ddb598a018d13cf2c4/include/crc32calc.h#L16 ?

src/mpegts/psi.h
```diff
@@ -5,6 +5,8 @@
 #include <variant>
 #include <vector>
 #include "BufferReader.h"
+#include "BufferWritter.h"
+#include "mpegts.h"
```
murillo128, line 9: do we need this include here?

murillo128 (changes requested):
almost lgtm, minor changes required though

src/mpegts/mpegtscrc32.cpp
harryz2000: Yes. This one is different. I tried the existing ones and they didn't work.

src/mpegts/psi.h
harryz2000: No need. removed.

src/mpegts/Packetizer.h
harryz2000: done

src/FrameDelayCalculator.cpp
harryz2000: OK. This is due to the PR I created was actually merging to this branch rather than the master. I can cherry pick the commit to master and do the rebase.

src/mpegts/mpegts.h
```diff
@@ -123,30 +146,41 @@ struct Packet
...
+	bool     originality = false;
+	bool     home = false;
+	bool     copyright = false;
+	bool     copyrightStart = false;
+	uint16_t frameLength = 0;
+	uint16_t bufferFullness = 0;
+	uint8_t  numberOfFramesMinus1 = 0;
+	uint16_t crc = 0;
+
+	// Serializable overrides
```
murillo128, line 176: remove outdated comments please

src/mpegts/Serializable.h
```diff
@@ -0,0 +1,25 @@
+#ifndef SERIALIZABLE_H
+#define SERIALIZABLE_H
```
murillo128, line 2: are we using this interface anymore? if not, please remove

harryz2000: It is only used in tests. Moved there for convenience.

harryz2000:
Will merge this shortly as per discussion.

PR #356, opened 2024-10-03
Fix mp3 decoding and add unit test

amaha877:


src/AudioDecoderWorker.cpp
```diff
@@ -164,13 +165,25 @@ int AudioDecoderWorker::Decode()
 					//Skip
 					continue;
 
-				// If it is aac and we have config
-				if (audioDecoder->type==AudioCodec::AAC && frame->HasCodecConfig())
+				if (frame->HasCodecConfig())
```
murillo128, line 168: as we pass now the full frame to the decoder, could we move the call the SetConfig to the decoder itself?

murillo128 (changes requested):

src/AudioDecoderWorker.cpp
amaha877: Not sure if i follow completely. Isn't that what we're doing with
`auto mp3 = static_cast<MP3Decoder*>(audioDecoder.get());
//Set config there
						mp3->SetConfig(frame->GetCodecConfigData(), frame->GetCodecConfigSize());`

murillo128: we are doing it on the worker, my suggestion was to move it to inside the Mp3/AacDecoder::Decode()  method instead.

yezhan10: if we want to use config that is stored in audio frame, guess a change is required to make in MPEGTSMPEG1AudioStream and MPEGTSMPEG2AudioStream:
from 	`frame.SetCodecConfig(reader.PeekData(), 4);` to ` frame.SetCodecConfig(mp3.PeekData(), 4);`
Also if possible, we also need to convert the clockrate from 90000 to sample rate of incoming audio.

amaha877: Sounds good. I moved the `setConfig()` call inside the decoders. We can address the mp3 config parsing in another PR.

src/mp3/MP3Decoder.cpp
```diff
@@ -76,9 +76,12 @@ bool MP3Decoder::SetConfig(const uint8_t* data,const size_t size)
 int MP3Decoder::Decode(const AudioFrame::const_shared& audioFrame)
 {
 	//Check we have config
-	if (!inited)
-		return Error("-AACDecoder::Decode() Not inited\n");
+	if (audioFrame->HasCodecConfig())
+		SetConfig(audioFrame->GetCodecConfigData(), audioFrame->GetCodecConfigSize());
```
yezhan10, line 80: Should we move this setConfig() to private member function if we intend to only allow decoder to modify it's config data?

amaha877: I thought about that but figured it would be better to keep it consistent with AACDecoder because the AACDecoder offers a `void SetAACConfig(const uint8_t* data,const size_t size);` public method that is exposed via SWIG in the audioCodecsNode. Perhaps someday we might need it exposed for MP3 as well?

yezhan10: yeah it makes sense, i got impression that this SetAACConfig() is never being used even though AudioDecoder.js has a function uses this function, but not sure what's it for. For consistency, guess we can leave it as it is for now.

murillo128: aac allows signaling config out of band (via sdp for example), that's why it is public.

murillo128 (approved):
lgtm

PR #358, opened 2024-10-10
mp3 config decode and bug fix

yezhan10:


src/mp3/MP3Config.h
```diff
@@ -0,0 +1,137 @@
...
+		MPEGVersion2,
+		MPEGVersion1,
+		MPEGVersionReserved
+	};
+	enum class MPEGAudioLayer {
+		MPEGLayer3,
+		MPEGLayerReserved
+	};
+
+	std::array<DWORD, 3> mpeg1Rates = {44100, 48000, 32000};
```
amaha877, line 21: would be more clear if we called it `mpeg1SamplingRates`, same for mpeg2 as well

src/mp3/MP3Config.h
```diff
@@ -0,0 +1,137 @@
...
+		if (size != 4) 
+			return Error("-MP3Config::Decode() config data must be size of 4 bytes\n");
+		
+		//Create bit reader
+		BufferReader reader(data,size);
+		BitReader r(reader);
+
+		try{
+			auto frameSync = r.Get(11);
+            if (frameSync != 0x7FF)
```
amaha877, line 46: I believe the `frameSync` can be 12 bits also? The doc says 

> "The first twelve bits (or first eleven bits in the case of the MPEG 2.5 extension) of a frame header are always set to 1"

When I use ffmpeg to test, I receive the header bytes as `0xff 0xfb 0x74 0x64` which will error out in this case

yezhan10: > MPEG Version 2.5 is not official standard. Bit No 20 in frame header is used to indicate version 2.5. Applications that do not support this MPEG version expect this bit always to be set, meaning that frame sync is twelve bits long, not eleven as stated here. Accordingly, mpegversion is one bit long (represents only bit No 19). I recommend using methodology presented here, since this allows you to distinguish all three versions and keep full compatibility.
 So i guess from parsing point of view, we can assume first 11 bits have to set, and use next two bits to represent version? 

> When I use ffmpeg to test, I receive the header bytes as 0xff 0xfb 0x74 0x64 which will error out in this case
The hex represents mpeg1, 48k sampling rate, stereo, and 96k bitrate, it's valid config data. It might have something to do the decoder was not able to deal with stereo since we did not set channel info in the mp3 decoder context in `SetConfig()`.

amaha877: Okay, makes sense.

amaha877:
lgtm

src/AudioTransrater.cpp
```diff
@@ -87,12 +85,19 @@ AudioBuffer::shared AudioTransrater::ProcessBuffer(const AudioBuffer::shared& au
...
-	resampledBuffer->SetTimestamp(scaledPTS);
+	resampledBuffer->SetTimestamp(scaleTimestamp(audioBuffer));
 	resampledBuffer->SetClockRate(outputRate);
 	//OK
 	return resampledBuffer;
+}
+
+uint64_t AudioTransrater::scaleTimestamp(const AudioBuffer::shared& audioBuffer)
+{
+	if (!playPTSOffset.has_value())
```
murillo128, line 96: I am not sure if I lile to set modify the `AudioTransrater` object on a method that is called `scaleTimestamp` which name doesn't imply side effects

src/mp3/MP3Config.h
murillo128: make it static and CamleCase

src/mp3/MP3Config.h
```diff
@@ -0,0 +1,137 @@
...
+	};
+	enum class MPEGAudioLayer {
+		MPEGLayer3,
+		MPEGLayerReserved
+	};
+
+	std::array<DWORD, 3> mpeg1Rates = {44100, 48000, 32000};
+	std::array<DWORD, 16> mpeg1Bitrates = {0, 32, 40, 48, 56, 64, 80, 96, 112, 128, 160, 192, 224, 256, 320, UINT_MAX};
+
+	std::array<DWORD, 3> mpeg2Rates = {22050, 24000, 16000};
```
murillo128, line 24: why we need the last UINT_MAX value?

yezhan10: no use, will remove

murillo128 (approved):
lgtm

PR #359, opened 2024-10-14
Don't reset running flag when exiting thread

harryz2000:
The reason is running flag was used for Stop() to determine whether need to join the thread. The thread must be joined before the thread object is freed.

### DIFF
--- a/src/EventLoop.cpp
+++ b/src/EventLoop.cpp
@@ -505,7 +505,7 @@ void EventLoop::Run(const std::chrono::milliseconds &duration)
 	auto until = duration < std::chrono::milliseconds::max() ? now + duration : std::chrono::milliseconds::max();
 		
 	//Run until ended
-	while(running && now<=until)
+	while(running && !exitCode.has_value() && now<=until)
 	{
 		//TRACE_EVENT("eventloop", "EventLoop::Run::Iteration");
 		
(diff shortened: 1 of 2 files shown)


src/EventLoop.cpp
```diff
@@ -505,7 +505,7 @@ void EventLoop::Run(const std::chrono::milliseconds &duration)
 	auto until = duration < std::chrono::milliseconds::max() ? now + duration : std::chrono::milliseconds::max();
 		
 	//Run until ended
-	while(running && now<=until)
+	while(running && !exitCode.has_value() && now<=until)
```
bcostdolby, line 508: We probably dont need "running" and exitCode. I.e. If exitCode has a value then the thread is not running right?

harryz2000: exitCode and running are flags to make the while loop to exit. Either of them is set, the while loop would exit, then the thread would exit. The running flag should only be used by EventLoop::Stop() function, as the thread is always required to be joined if it is joinable. So for stopping caused inside while loop, I used the exitCode, which is an optional, which is only be set when thread needs to be stopped.

harryz2000: To answer your question, I have changed the code exitCode can have a value while the thread is running, it would cause the thread to stop/exit.

PR #361, opened 2024-10-17
fix: update audio pts conversion calculation to avoid overflow

yezhan10:
this PR is to update the pts conversion function in audio transrater, and remove unused header from AudioDecoderWorker.h.


amaha877 (approved):
lgtm

PR #363, opened 2024-10-21
Use minus 1 instead of errno to indicate system poll error

harryz2000:


src/SystemPoll.cpp
```diff
@@ -142,7 +142,8 @@ std::pair<uint16_t, int> SystemPoll::GetEvents(int fd) const
 	
 	if ((revents & POLLHUP) || (revents & POLLERR))
 	{
-		return std::make_pair<>(0, errno);
+		Error("-SystemPoll::GetEvents() | Error events: 0x%x errno: %d\n", revents, errno);
```
bcostdolby, line 145: I feel like we should add the FD number to these logs (below as well) as well so we might be able to match it up with what it is failing for later.

bcostdolby (changes requested):

src/SystemPoll.cpp
harryz2000: Done.

bcostdolby (approved):
lgtm adding bigger change later in DIOS-7350

PR #367, opened 2024-10-23
split nalu to equal rtp packet size

yezhan10:


src/h264/H26xPacketizer.cpp
```diff
@@ -54,22 +53,23 @@ void H26xPacketizer::EmitNal(VideoFrame& frame, BufferReader nal, std::string& f
 	//Skip payload nal header
 	nal.Skip(naluHeaderSize);
 	pos += naluHeaderSize;
-	//Split it
-	while (nal.GetLeft())
+	
+	auto numPackets = nal.GetLeft() / (RTPPAYLOADSIZE - fuPrefixSize) + 1;
```
murillo128, line 57: wouldn't this give an incorrect packet count if mod is 0? i.e. if `nal.GetLeft()` is `(RTPPAYLOADSIZE - fuPrefixSize)`

`auto numPackets = (RTPPAYLOADSIZE - fuPrefixSize) / (RTPPAYLOADSIZE - fuPrefixSize) + 1; ` is 2 an not 1.


```suggestion
	auto numPackets = std::ceil(nal.GetLeft() / (RTPPAYLOADSIZE - fuPrefixSize));
```

murillo128: could you add a test for that as well?

murillo128 (changes requested):

src/h264/H26xPacketizer.cpp
yezhan10: if `nal.GetLeft()` is `(RTPPAYLOADSIZE - fuPrefixSize)` then it can be fit into one single packet right? which would not need FU-A anymore I think. I can definitely add a test with nalUnitSize == 1200 bytes which is the min bytes for FU-A mode.

murillo128: what I mean is that the code above will fail if mod=0

test/unit/TestH26xRTPPacketization.cpp
```diff
@@ -0,0 +1,292 @@
...
+        BYTE nalLID =  (naluHeader >> 3) & 0b111111;
+        BYTE nalTID = naluHeader & 0b111;
+
+        BufferReader nal(nalData, nalSize);
+        VideoFrame frame(VideoCodec::H265);
+        h265Packetizer.TestRTPFUA(frame, nal);
+
+        int expectedRTPPacekts = std::ceil((double)nalPayloadSize / (RTPPAYLOADSIZE - FUPayloadHdrSize - FUHeaderSize));
+        auto packetLen = nalPayloadSize / expectedRTPPacekts;
+        int mod = nalPayloadSize % expectedRTPPacekts;
```
murillo128, line 219: please check `mod == 0`

src/h264/H26xPacketizer.cpp
```diff
@@ -54,22 +53,22 @@ void H26xPacketizer::EmitNal(VideoFrame& frame, BufferReader nal, std::string& f
 	//Skip payload nal header
 	nal.Skip(naluHeaderSize);
 	pos += naluHeaderSize;
-	//Split it
-	while (nal.GetLeft())
+	uint32_t numPackets = std::ceil(static_cast<double>(nal.GetLeft()) / (RTPPAYLOADSIZE - fuPrefixSize));
+	auto packetLen = nal.GetLeft() / numPackets;
```
harryz2000, line 57: should we check if numPackets == 0?

src/h264/H26xPacketizer.cpp
```diff
@@ -54,22 +53,22 @@ void H26xPacketizer::EmitNal(VideoFrame& frame, BufferReader nal, std::string& f
 	//Skip payload nal header
 	nal.Skip(naluHeaderSize);
 	pos += naluHeaderSize;
-	//Split it
-	while (nal.GetLeft())
+	uint32_t numPackets = std::ceil(static_cast<double>(nal.GetLeft()) / (RTPPAYLOADSIZE - fuPrefixSize));
```
harryz2000, line 56: Looks if mod != 0, one  more bytes will be added to each packet. Should we change this to?
```suggestion
	uint32_t numPackets = std::ceil(static_cast<double>(nal.GetLeft()) / (RTPPAYLOADSIZE - fuPrefixSize - 1));
```

harryz2000: OK. Probably not an issue as you calculate packet len later.

test/unit/TestH26xRTPPacketization.cpp
```diff
@@ -0,0 +1,298 @@
...
+        VideoFrame frame(VideoCodec::H264);
+        BufferReader nal(nalData, nalSize);
+
+        h264Packetizer.TestRTPFUA(frame, nal);
+
+        int expectedPos = 4 + H264NalHeaderSize;
+        int expectedRTPPacekts = std::ceil((double)nalPayloadSize / (RTPPAYLOADSIZE - FUPayloadHdrSize - FUHeaderSize));
+        auto packetLen = nalPayloadSize / expectedRTPPacekts;
+        int mod = nalPayloadSize % expectedRTPPacekts;
+        EXPECT_TRUE(mod!=0);
```
harryz2000, line 46: Can we change EXPECT_* to ASSERT_*? Mostly we shouldn't need to continue to run tests if previous check failed. Otherwise sometimes it might cause crash.

harryz2000, line 46: Consider use:
```suggestion
        ASSERT_TRUE(0, mod);
```
It would give more useful info if it fails.

src/h264/H26xPacketizer.cpp
```diff
@@ -54,22 +53,22 @@ void H26xPacketizer::EmitNal(VideoFrame& frame, BufferReader nal, std::string& f
...
+	auto packetLen = nal.GetLeft() / numPackets;
+	int mod = nal.GetLeft() % numPackets;
```
harryz2000, line 58: I think it would be good if we have test cases for following scenarios.
1. mod == 1
2. mod == (packetLen - 1)
3. mod == 0
You probably have added. Just wanted to confirm as I am not sure value of each constants.

src/h264/H26xPacketizer.cpp
yezhan10: i dont think it's required, since we only split nalu if the size is greater than 1200 bytes

murillo128 (approved):
lgtm

PR #368, opened 2024-10-24
DIOS-7446 Avoid calling pthread_kill if thread is not running

harryz2000:


murillo128 (approved):
lgtm

src/rtmp/rtmpconnection.cpp
```diff
@@ -311,7 +311,8 @@ void RTMPConnection::SignalWriteNeeded()
 	pthread_mutex_unlock(&mutex);
 
 	//Signal the pthread this will cause the poll call to exit
-	pthread_kill(thread.native_handle(),SIGIO);
+	if (running)
```
craig-johnston, line 314: I don't know this code, so general question:
Is there any chance of a race condition here? Can the thread stop running while this code (the next line specifically) is executing?
The variable running isn't being read under a mutex.

craig-johnston (changes requested):

src/rtmp/rtmpconnection.cpp
harryz2000: Yeah. Ideally, we would use EventLoop which has the eventfd to break the poll wait. This is a hot fix.

craig-johnston: It'll definitely crash a lot less and if the better fix is coming later, all good.

PR #370, opened 2024-10-25
Dev/amaha/dios-7395/fix srt hevc publish

amaha877:
This PR fixes [DIOS-7395](https://jira.dolby.net/jira/browse/DIOS-7395)
as well as adds unit tests for SPS, PPS and VPS parsing in H265. Note that our parser currently doesn't fully flesh out all the fields currently. Added a backlog [story](https://jira.dolby.net/jira/browse/DIOS-7468) for that

### DIFF
--- a/src/h264/H26xNal.h
+++ b/src/h264/H26xNal.h
@@ -61,7 +61,7 @@ public:
 			}
 			if (len == num)
 				return true;
-		}
+			}
 		return false;
 	}
 
@@ -98,7 +98,7 @@ public:
 			}
 			if (len == num)
 				return;
-		}
+			}
 
 		throw std::range_error("couldn't skip more bytes");
 	}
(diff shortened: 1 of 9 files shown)


src/h264/H26xNal.h
```diff
@@ -133,7 +133,10 @@ class RbspReader
 				len++;
 			}
 			if (len == num)
+			{
```
yezhan10, line 136: do we have to set number of zeros in the Skip() function (line 99)?

murillo128 (approved):
lgtm

src/h264/H26xNal.h
amaha877: Yeah I thought about that a lot. We don't need to do it here and the reason is due to the call sequences to these functions. For example, to load the cache we first `Assert(num_bytes)` and then we use `Peek(num_bytes)` to load the cache. So we only need to update `zeros` when we assert. 
Similarly, the `Skip()` function in the `RbspReader` is only called by `SkipCached` which has already accounted for the emulation prevention byte. Happy to discuss more offline, hence had to write unit tests to keep the fix honest :)

yezhan10: Thanks for the detailed explanation.

PR #372, opened 2024-10-31
split onMediaFrame function

yezhan10:
currently onMediaFrame does lots of things, would be better to create separate functions for tasks like packetization and media stats update.


src/MediaFrameListenerBridge.cpp
```diff
@@ -503,16 +392,78 @@ void  MediaFrameListenerBridge::onMediaFrameAsync(std::chrono::milliseconds now,
...
 		//Insert it
 		packets.emplace(scheduled, packet, std::chrono::milliseconds(packetDuration));
 
 		//Recalcualte pending
 		pendingLength -= info[i].GetTotalLength();
 		pendingDuration -= packetDuration;
+	}
+}
+
+void MediaFrameListenerBridge::UpdateAccumulatorStats(const std::shared_ptr<MediaFrame>& frame, std::chrono::milliseconds now)
```
harryz2000, line 404: I would prefer not change this in this PR as the value calculated in this function are not only for stats, for example the pendingDuration, etc. Although it might make the original function smaller, it might make the code difficult to read as variables are set in different places.

If this function is to update stats, imo it should only update stats, and no others. However, I think just leaveing these code in the original function is fine.

include/MediaFrameListenerBridge.h
```diff
@@ -120,6 +120,8 @@ class MediaFrameListenerBridge :
 
 	MediaFrame::Type type = MediaFrame::Type::Unknown;
 	BYTE codec = -1;
+	DWORD rate = 1000;
+	DWORD pendingDuration = 0;
```
harryz2000, line 124: As the other comment, I think it would be good to keep these variables local in the original function.

src/MediaFrameListenerBridge.cpp
yezhan10: how about we have a standalone function to update stats, and i can remove the pendingDuration from this function?

yezhan10: The reason that I created such standalone function is to avoid the mediaframe cast to videoframe and audioframe at multiple times. Any suggestions to avoid this in a better way?

harryz2000: I would suggest to split to function by functionality. My preference would be:
OnMediaFrameAsync() is the parent function and manages the dispatch and scheduling.

Create following functions:
1. CreateRTPPacket()
     The input of this function would be the frame and the rtppacketinfo. CreateRTPPacket() should try to fill the rtp payload size as much as possible. The return value should be the index of the next unconsumed rtppacketinfo. So the caller code can advance to create another packet, i.e.

```
std::pair<size_t, std::shared_ptr<RTPPacket>> CreateRTPPacket(const std::shared_ptr<MediaFrame>& frame, const MediaFrame::RtpPacketizationInfo& info, size_t startIndex)
{
...
}

```
  OnMediaFrameAsync should call CreateRTPPacket() in a loop until the index reaches the end of the RtpPacketizationInfo.

2. UpdateStats()
This function should only update stats and do nothing else. It should update all the stats, including the `waited`, etc.

yezhan10:
> please, do not do code extraction and refactor on the same PR as it makes reviewing much mor difficult. If you want to extract code to a different fucntion, do a PR to move the code verbatim, and then a follow up PR to refactor the code.

I have updated this PR again. The only "refactor" i am aware of now is that adding two member variables to this class, it's necessary IMO if i split the onMediaFrame into packetization and updateFrameInfoAndStats. I really want to avoid casting the MediaFrame to AudioFrame or VideoFrame multiple times inside the function.

harryz2000:
I created a commit to describe what I meant. It is not tidied and not tested.
https://github.com/medooze/media-server/commit/9fffea20ae9484a4ed46b158b24a83be716887f1

I don't really mind of the frame casting, although that would be nice. I didn't look very deep to remove the duplicated masking. But I think it is more important to make function to have single functionality.

yezhan10:
>But I think it is more important to make function to have single functionality.

Thanks Harry for taking your time to review and even have an example. I would prefer that a for loop inside CreatePacket() instead of a for loop in OnMediaFrame(). Also I guess your major concern is UpdateFrameInfoAndStats() does two things instead one thing? If this is the case, I can split into two functions. One called UpdateFrameInfo and the other called UpdateMediaStats()? IMO storing some frame related variables such as pendingDuration and frameRate would not be problem at all? We did have to use those variables in various places, storing it as a member variable makes sense to me.

harryz2000:
>  I would prefer that a for loop inside CreatePacket() instead of a for loop in OnMediaFrame()
My impression this refactor is to make it easy to combine NALUs into one RTP packet. Let me if my understanding is wrong.
In the commit I made,  CreateRTPPacket() function is meant to create a single RTP packet. it just consume one RTPInfo. In next step to combine NALUs, CreateRTPPacket() may consume multiple RTPInfo (NALUs) to create a single RTP packet, in which case a loop will be needed inside CreateRTPPacket(). The outside loop in OnMediaFrameAsync() is for each packet. The loop inside CreateRTPPacket() would be for each NALU.

yezhan10:
> My impression this refactor is to make it easy to combine NALUs into one RTP packet. Let me know if my understanding is wrong.

Not really, this PR as suggested by Sergio for easier review experience, is only about split onMediaFrame based on functionalities so that OnMediaFrame() itself isn't this lengthy ( Now its around ~400 line?). Once this is approved, I will have another PR.

harryz2000:
That is right. The change I was suggesting wouldn't change the current logic. It is just restructuring. It prepares the code for next change though.

src/MediaFrameListenerBridge.cpp
```diff
@@ -571,6 +475,77 @@ void  MediaFrameListenerBridge::onMediaFrameAsync(std::chrono::milliseconds now,
 	}
 }
 
+void MediaFrameListenerBridge::UpdateFrameInfoAndStats(const std::shared_ptr<MediaFrame>& frame, std::chrono::milliseconds now, std::chrono::milliseconds scheduled)
+{
+	type = frame->GetType();
+	uint32_t frameSize = 0;
```
murillo128, line 481:
```suggestion
	uint32_t frameSize = audioFrame->GetLength();
```

src/MediaFrameListenerBridge.cpp
```diff
@@ -571,6 +475,77 @@ void  MediaFrameListenerBridge::onMediaFrameAsync(std::chrono::milliseconds now,
...
+	uint64_t ms = scheduled.count();
+	switch(type)
+	{
+		case MediaFrame::Audio:
+		{
+			//get audio frame
+			auto audioFrame = std::static_pointer_cast<AudioFrame>(frame);
+			// Note: This may truncate UNKNOWN but we do that many places elsewhere treating -1 == 0xFF == UNKNOWN so being consistent here as well
+			codec = audioFrame->GetCodec();
+			frameSize = audioFrame->GetLength();
```
murillo128, line 492:
```suggestion
```

src/MediaFrameListenerBridge.cpp
```diff
@@ -571,6 +475,77 @@ void  MediaFrameListenerBridge::onMediaFrameAsync(std::chrono::milliseconds now,
...
+		}
+		case MediaFrame::Video:
+		{
+			//get Video frame
+			auto videoFrame = std::static_pointer_cast<VideoFrame>(frame);
+			// Note: This may truncate UNKNOWN but we do that many places elsewhere treating -1 == 0xFF == UNKNOWN so being consistent here as well
+			codec = videoFrame->GetCodec();
+			//Set clock rate
+			rate = 90000;
+			frameSize = videoFrame->GetLength();
```
murillo128, line 506:
```suggestion
```

src/MediaFrameListenerBridge.cpp
```diff
@@ -272,120 +272,15 @@ void  MediaFrameListenerBridge::onMediaFrameAsync(std::chrono::milliseconds now,
...
-	acumulator.Update(now.count(),frameSize);
-
-	//Get scheduled time in ms
-	uint64_t ms = scheduled.count();
-	//Waiting time uses scheduled time
-	waited.Update(ms, ms>time ? ms-time : 0);
-	//Get bitrate in bps
-	bitrate = acumulator.GetInstant()*8;
-
+	UpdateFrameInfoAndStats(frame, now, scheduled);
```
murillo128, line 283: let's split this function in one getting frame info and another one updating the stats

src/MediaFrameListenerBridge.cpp
```diff
@@ -402,6 +297,15 @@ void  MediaFrameListenerBridge::onMediaFrameAsync(std::chrono::milliseconds now,
 		firstTimestamp = frame->GetTimeStamp();
 		Log("-MediaFrameListenerBridge[%p]::onMediaFrame Handling frame: %llu Resetting firstTimestamp to: %llu\n", this, frame->GetTimestamp(), firstTimestamp);
 	}
+	
+	Packetize(frame, scheduled);
+}
+
+void MediaFrameListenerBridge::Packetize(const std::shared_ptr<MediaFrame>& frame, std::chrono::milliseconds scheduled)
+{
+	const MediaFrame::RtpPacketizationInfo& info = frame->GetRtpPacketizationInfo();
```
murillo128, line 306: I think it would be helpfull to extract the rtp packetization code from this class and maybe move it to an static method in the RTPPacketizer class:

```
  std::vector<RTPPacket::shared> RTPPacketizer::GenerateRTPPackets(frame);
```
```

murillo128 (approved):
lgtm, left feedback for the upcoming refactor actually doing the changes

PR #373, opened 2024-10-31
aggregate h264 packets into a single stap-a packet

murillo128:


src/MediaFrameListenerBridge.cpp
```diff
@@ -423,13 +423,69 @@ void  MediaFrameListenerBridge::onMediaFrameAsync(std::chrono::milliseconds now,
 		if (rtp.GetTotalLength()>packet->GetMaxMediaLength())
 			//Error
 			continue;
+
+		//Try to aggreate multiple NALs into a single STAP-A packet for H264
+                if (frame->GetType() == MediaFrame::Video && codec == VideoCodec::H264)
```
yezhan10, line 428: maybe add additional conditions, such as if the packet is not the last one and no prefix.

murillo128: that's done below in line 440 (as it checks current element)

PR #374, opened 2024-11-01
add STAP-A aggregation for h264

yezhan10:


include/MediaFrameListenerBridge.h
```diff
@@ -97,7 +97,7 @@ class MediaFrameListenerBridge :
 	DWORD extSeqNum = 0;
 	bool  smooth = true;
 	std::set<RTPIncomingMediaStream::Listener*> listeners;
-        std::set<MediaFrame::Listener::shared> mediaFrameListeners;
+    std::set<MediaFrame::Listener::shared> mediaFrameListeners;
```
bcostdolby, line 102: nitpick. Something wrong with the whitespace

src/MediaFrameListenerBridge.cpp
```diff
@@ -319,79 +319,55 @@ void MediaFrameListenerBridge::Packetize(const std::shared_ptr<MediaFrame>& fram
...
+					uint8_t nalLen[StapANaluSize] = {};
+					//Set aggregation length
+					set2(nalLen, 0, info[aggIdx].GetSize());
+					//Set it
+					packet->AppendPayload(nalLen, sizeof(nalLen));
+					//Set data
+					packet->AppendPayload(frameData + info[aggIdx].GetPos(), info[aggIdx].GetSize());
+				}
+				i = stapAEndIdx-1;
+				bool isLast = i==info.size()-1 ? true : false;
```
bcostdolby, line 355: Please move there to early in loop so not duplicated here and in else

src/MediaFrameListenerBridge.cpp
```diff
@@ -319,79 +319,55 @@ void MediaFrameListenerBridge::Packetize(const std::shared_ptr<MediaFrame>& fram
...
-			packet->SetMark(false);
-			
-		//Increase stats
-		numPackets++;
-		//Set last sent time
-		lastTime = getTime();
-
-		//Fill payload descriptors
-		if (frame->GetType()==MediaFrame::Video)
+		if (frame->GetType() == MediaFrame::Video && codec == VideoCodec::H264 && info.size()>i+1 && rtp.GetPrefixLen()==0)
```
bcostdolby, line 327: Could we put this use case in a separate function to clean it up a little?

src/MediaFrameListenerBridge.cpp
```diff
@@ -319,79 +319,55 @@ void MediaFrameListenerBridge::Packetize(const std::shared_ptr<MediaFrame>& fram
...
-
-			//TODO: move out of here
-			VideoLayerSelector::GetLayerIds(packet);
-				
-			// Set frame resolution if needed
-			if ((video->GetWidth() == 0 || video->GetHeight() == 0) && 
-				(packet->GetWidth() != 0 && packet->GetHeight() != 0))
+			auto stapAEndIdx = i;
+			auto currStapALen = StapAHeaderSize;
+			while (stapAEndIdx < info.size() && info[stapAEndIdx].GetPrefixLen()==0)
```
bcostdolby, line 331: Might be helpful to have a comment here. I can read the code but not sure why/when we might have a prefix and what that means?

src/MediaFrameListenerBridge.cpp
```diff
@@ -319,79 +319,55 @@ void MediaFrameListenerBridge::Packetize(const std::shared_ptr<MediaFrame>& fram
...
 			{
-				width = packet->GetWidth();
-				height = packet->GetHeight();
+				if (currStapALen > packet->GetMaxMediaLength())
```
bcostdolby, line 333: nitpick: Style wise I think this might be better as a for loop or at least a larger conditional in the while loop. 

The condition is split in multiple places.

Looks like we are looping over all packets while no prefix and size is under the limit. 

Is the goal here to calc a specific index? If so maybe this can belong in a separate function.

src/MediaFrameListenerBridge.cpp
```diff
@@ -319,79 +319,55 @@ void MediaFrameListenerBridge::Packetize(const std::shared_ptr<MediaFrame>& fram
...
 			{
-				int streamIdx = 0, spatialId = 0, numRtpStreams = 1;
-				//Create VLA info
-				VideoLayersAllocation::SpatialLayer spatialLayer(streamIdx, spatialId, std::vector<uint16_t>{ (uint16_t)targetBitrate }, video->GetWidth(), video->GetHeight(), video->GetTargetFps());
-				std::vector<VideoLayersAllocation::SpatialLayer> activeSpatialLayers{spatialLayer};
-				VideoLayersAllocation videoLayersAllocation = {streamIdx, numRtpStreams, activeSpatialLayers};
-											
-				//Set it in packet
-				packet->SetVideoLayersAllocation(videoLayersAllocation);
+				uint8_t stapAHeader = 0x78;
```
bcostdolby, line 341: Again I think this will be easier to read if the aggregation loop is in a function. Maybe even the loop inside the function?

Something like:
```
stapAEndIdx = CalculateAggregationEndIndex();
AggregatePayloads(stapAEndIdx, ...)

```

src/rtp/RTPPayload.cpp
```diff
@@ -56,6 +56,19 @@ bool RTPPayload::PrefixPayload(BYTE *data,DWORD size)
 	return true;
 }
 
+bool RTPPayload::AppendPayload(const BYTE* data, DWORD size)
+{
+	//Check size
+	if ((payload - buffer.data ()) + payloadLen + size > GetMaxMediaLength ())
```
bcostdolby, line 62: Maybe another function to get payload prefix buffer size instead of just payload - buffer.data()

Also I am just skimming over this, but something is confusing about this conditional to me. 

payload - buffer.data() : Is this just always == PREFIX?

Also GetMaxMediaLength() is returning SIZE not sure why we need to consider the prefix here...

Can we just do 
`if (payloadLen + size) > GetMaxMediaLength()`

or is the max also supposed to include the prefix despite the buffer looking like it is otherwise?

bcostdolby:
Is is possible to do the refactor of moving the code into separate functions in a separate branch?

Its just a little more complex to review the changes when things are both changed and moved around.I reviewed the commits one at a time to get around this but nice for merge history to look separate

src/rtp/RTPPayload.cpp
yezhan10: yeah it makes more sense to me to do, i will update the code
`if (payloadLen + size) > GetMaxMediaLength()`

PR #375, opened 2024-11-01
DIOS-7462 Avoid FileDescriptor released twice for same fd

harryz2000:


bcostdolby (approved):
lgtm

mildsunrise:
I'm not following, how was the previous code releasing the same FD twice?

bcostdolby:
Using the dup2() would work as well yes but I would argue the dup2() is adding more unnecessary overhead (not necessarily performance but extra unnecessary FDs used). Conceptually the shared ptr makes more sense since we really only have a single descriptor and just want to label it differently. Both would work.

PR #376, opened 2024-11-01
DIOS-7454  Fixed Safari playback issue for duplicated parameter sets (happens when we re-packetize a video frame)

harryz2000:


src/rtmp/rtmppacketizer.cpp
```diff
@@ -331,13 +326,47 @@ std::unique_ptr<VideoFrame> RTMPH26xPacketizer<DescClass, SPSClass, PPSClass, co
...
+			auto naluHeader = get2(nal, 0);
+			BYTE nalUnitType = (naluHeader >> 9) & 0b111111;
+			
+			switch (nalUnitType)
 			{
-				auto naluHeader = get2(nal, 0);
-				BYTE nalUnitType = (naluHeader >> 9) & 0b111111;
-				if ((nalUnitType <= HEVC_RTP_NALU_Type::RASL_R) || 
-					(nalUnitType >= HEVC_RTP_NALU_Type::BLA_W_LP && nalUnitType <= HEVC_RTP_NALU_Type::CRA_NUT))
+			case HEVC_RTP_NALU_Type::TRAIL_N:
```
harryz2000, line 353: A note here. I think we would prefer to use [Pascal case]( https://khalilstemmler.com/blogs/camel-case-snake-case-pascal-case/#Pascal-case) for constants and enums. I didn't change as these are existing code. Happy to change if it is preferred.

src/h264/h264.h
```diff
@@ -41,6 +41,41 @@ static constexpr uint8_t  Log2MaxMvLengthVerticalMin = 0;
 static constexpr uint8_t  Log2MaxMvLengthVerticalMax = 16;
 static constexpr uint8_t  MaxDpbFrames = 16; // Lifted from FFMPEG
 
+enum AvcNaluType {
```
murillo128, line 44:
```suggestion
enum class AvcNaluType {
```

src/h264/h264.h
```diff
@@ -41,6 +41,41 @@ static constexpr uint8_t  Log2MaxMvLengthVerticalMin = 0;
...
+	AvcNaluTypeUnspecified		= 0,
```
murillo128, line 45: remove `AvcNalu` as we are making the `enum` and `enum class` we will access it by `AvcNaluType::Unspecified`

src/rtmp/rtmppacketizer.cpp
murillo128: lets change at least the enum name in a later PR

src/rtmp/rtmppacketizer.cpp
```diff
@@ -331,13 +326,47 @@ std::unique_ptr<VideoFrame> RTMPH26xPacketizer<DescClass, SPSClass, PPSClass, co
 						parsingFrameType = false;
 					}
 				}
+				break;
+			case AvcNaluTypeSps:
+				// Skip if we already have sps. Safari may complain about duplicated parameter sets.
```
murillo128, line 331:
```suggestion
				// Skip if we already have sps inserted from the AVCDescriptor
				// Safari may complain about duplicated parameter sets.
```

murillo128 (approved):
lgtm

PR #377, opened 2024-11-05
refactor: proper abstraction of event fd for apple/linux

murillo128:


include/PollSignalling.h
```diff
@@ -5,24 +5,35 @@
 #include <atomic>
 #include <memory>
 
+class EventPipe
+{
+public:
+	inline FileDescriptor& GetReadFd() const
+	inline FileDescriptor& GetWriteFd() const
```
mildsunrise, line 12: it's not very useful to declare something `inline` if we don't give its definition in the .h, could we move it here?
also missing `const` and semicolon
```suggestion
#if __APPLE__
	const FileDescriptor& GetReadFd() const { return pipeFds[0]; }
	const FileDescriptor& GetWriteFd() const { return pipeFds[1]; }
#else
	const FileDescriptor& GetReadFd() const { return eventFd; }
	const FileDescriptor& GetWriteFd() const { return eventFd; }
#endif
```

src/PollSignalling.cpp
```diff
@@ -4,50 +4,59 @@
...
-	fcntl(pipe[1], F_SETFL , O_NONBLOCK);
-	
-	pipeFds[0] = std::make_shared<FileDescriptor>(pipe[0]);
-	pipeFds[1] = std::make_shared<FileDescriptor>(pipe[1]);
-	
+	fcntl(pipe[0], F_SETFL, O_NONBLOCK);
+	fcntl(pipe[1], F_SETFL, O_NONBLOCK);
+
+	pipeFds[0] = pipe[0];
+	pipeFds[1] = pipe[1];
```
mildsunrise, line 23: missing constructor call I think
```suggestion
	pipeFds[0] = FileDescriptor(pipe[0]);
	pipeFds[1] = FileDescriptor(pipe[1]);
```

mildsunrise (approved):
other than the suggestions, lgtm

include/PollSignalling.h
murillo128: I don't like having the actual methods definition inside `#if/#else/#endif` as we could have different class method and not just class implementations. Not sure if this pattern is better:

```
class EventPipe
{
public:
	inline FileDescriptor& GetReadFd () const
	{
#if __APPLE__
		return pipeFds[0];
#else
		return eventFd;
#end
	}
	inline FileDescriptor& GetWriteFd () const
	{
#if __APPLE__
		return pipeFds[1];
#else
		return eventFd;
#end
	}
private:
#if __APPLE__
	FileDescriptor pipeFds[2];
#else
	FileDescriptor eventFd;
#end
};
```
or if we could just remove the `inline` keyword completely. wdyt?

src/PollSignalling.cpp
murillo128: is it really needed? not sure either

murillo128: yes, no int operation conversion overload

include/PollSignalling.h
mildsunrise: your point is valid, and we can do that yeah.
`inline` does nothing when you supply the definition inline, as it's implicit in this case.
whatever you do, the method needs to return `const FileDescriptor&` if it's a const method, otherwise you'd be violating const safety

PR #379, opened 2024-11-05
fix(rtmp): only send SEI messages on Intra frames to avoid freezing on packet losses

murillo128:


amaha877:
Don't we need to verify if all these unregistered SEI messages are only sent on I-Frames?
https://confluence.dolby.net/kb/display/DLBCLS/H.264+SEI+unregistered+data+UUIDs

For example the SCTE data in `H264Packetizer.cpp` doesn't limit to Iframes and also repeats over the next 20 frames:
```
                EmitNal(frame, BufferReader(seiEscaped, seiSize));
		if (repeatFrameCounter-- <= 0)
		{
			scteMessage.Reset();
			repeatFrameCounter = scteFrameRepeatCount;
		}
```

But if its only concerning RTMP input then it should be ok.

mildsunrise:
@murillo128 lgtm

@amaha877 yeah, we should probably also do something about SCTE-35... this issue is regardless of ingest, see https://dolbyio.slack.com/archives/C047A9VSD9V/p1730826452266059

PR #382, opened 2024-11-07
refactor: RTMPConnection to use EventLoop to avoid race condition

harryz2000:
This is to use eventfd to signal thread, instead of pthread_kill and running flag, which could cause crashes due to race condition.


include/rtmp/rtmpconnection.h
```diff
@@ -128,8 +137,7 @@ class RTMPConnection :
 	DWORD maxChunkSize;
 	DWORD maxOutChunkSize;
 
-	std::thread thread;
-	pthread_mutex_t mutex;
+	std::mutex mutex;
```
murillo128, line 140: do we still need the mutex? can we change it by async calls?

include/rtmp/rtmpconnection.h
```diff
@@ -151,6 +159,11 @@ class RTMPConnection :
 	DWORD bandCalc = 0;
 	
 	DWORD rtt = 0;
+	
+	static const unsigned int CacheSize = 1400;
+	BYTE cache[CacheSize];
```
murillo128, line 164: I don't think it is needed to have it in an object attribute, can't this be a method variable instead?

murillo128 (commented):
aren't we loosing the RTMP timeout if no media is received in 30s in this change?

include/rtmp/rtmpconnection.h
harryz2000: We could. I will try. I was trying to not change too much logic.

include/rtmp/rtmpconnection.h
harryz2000: Previously this array was created in the run() function only once. We are now using callbacks. Using local variable may cause frequently allocate/deallocate as the OnPollIn() and OnPollOut() was called frequently. The array is big and only needed be created once, kind of a temporary buffer. I can try but not sure if there is a better way to not using object attribute and at same time avoid the allocation issue.

murillo128: I kind of always preffer stack vs heap, but I may be wrong here. @mildsunrise @bcostdolby wdyt?

harryz2000:
> aren't we loosing the RTMP timeout if no media is received in 30s in this change?

Correct. I would add the timeout back.

include/rtmp/rtmpconnection.h
harryz2000: Another way is to use thread_local. i.e. in `rtmpconnection.cpp`
```
namespace
{
static constexpr unsigned int CacheSize = 1400;
static thread_local BYTE cache[CacheSize];
}
```

However, I would prefer to use the object attribute as it is explicitly freed once the object is released. If we use thread_local, it would be no problem if we only use it in the RTMPConnection thread, but if we accidentally use it in a thread that may last for long time, it might cause memory leak.

harryz2000: @bcostdolby I changed the name "cache" to "buffer" as the array is not really for a cache. It is just a memory preallocated for using in the thread. It is nothing to do with other member variables.

include/rtmp/rtmpconnection.h
harryz2000: done

include/rtmp/rtmpconnection.h
bcostdolby: I understand it is just a buffer. I wasn't worried about the naming buffer or cache work fine. 

What I was proposing was a super minor optimization (more a nitpick really, but from the above we do seem to be caring a bit about the micro optimisation from the questions above about wether to place this on the stack or heap). 

Hardware cache lines are typically like 64 bytes. If a member variable "eventMask" is accessed regularly at about the same time other member variables are accessed in the same class, then they can make use of the fact there is a chance the two are loaded on the same hardware cache line and it is likely hot in cache being faster to access.

By placing it after the "buffer" then there is no chance that it could be on the same cache line so would need to load a new 64byte segment into ram before access almost every time. 

Just generally a good idea to place most frequently used member variable together in the class so the memory layout is such that they might be able to share the same hardware cache line. 

If you start adding buffers between them then it forces them out of the same cache line (useful for some scenarios like containers with known threads not wanting to share cache lines, but otherwise a bad idea generally).

src/EventLoop.cpp
```diff
@@ -241,8 +241,13 @@ bool EventLoop::Stop()
...
 
-		//Nulifi thread
-		thread.join();
+		//If we are on different thread
+		if (std::this_thread::get_id()!=thread.get_id())
+			//Join it
+			thread.join();
+		else
+			//Detach as we are ending ourselves
+			thread.detach();
```
bcostdolby, line 250: This sort of thing is *usually* an anti-pattern as you cant always shut down safely. 

I haven't looked at the rest of it in heaps of detail and it is unlikely a serious issue but if the main event loop can call stop on itself, then I assume we Signal() above, detach, then this thread will eventually go back out to the main loop, check the signal see it is stopped and then slowly exit itself?

If so after the detach but before it executes the main loop to check the signal again, if the server then asks to shutdown before the self signalled thread exits, then there is a chance the shutdown thread will think this thread is all cleaned up and in fact it isn't quite cleaned up yet. 

Not sure there are bugs here like can it delete the memory and cause a shutdown crash? 

Again unlikely to happen but usually calling detach() is a last resort anti-pattern as it can cause races.

include/EventLoop.h
```diff
@@ -149,6 +154,7 @@ class EventLoop : public TimeService
 
 	std::thread			thread;
 	std::unique_ptr<Poll>		poll;
+	int				defaultTimeoutMs = 10E3;
```
bcostdolby, line 157: In the past used more like 100msec for these timeouts instead of 10msec as they can add overhead waking up unnecessarily and a 100msec wait for something like this is usually not a problem.

src/SystemPoll.cpp
```diff
@@ -182,10 +182,16 @@ int SystemPoll::Wait(uint32_t timeOutMs)
...
 	
 	// Wait for events
-	if (poll(syspfds.data(), syspfds.size(),timeOutMs) < 0)
+	if (auto ret = poll(syspfds.data(), syspfds.size(),timeOutMs) <= 0)
 	{
-		Error("-SystemPoll::Wait() | poll() error. errno: %d\n", errno);
-		return -1;
+		if (ret < 0)
+		{
+			Error("-SystemPoll::Wait() | poll() error. errno: %d\n", errno);
```
bcostdolby, line 189: Maybe just add strerror here and others I hate having to lookup error codes :-)

include/rtmp/rtmpconnection.h
harryz2000: Thanks for the explanation. The eventMask is now above the buffer. Wouldn't that meet your point?

bcostdolby: Yep. I saw that but didn't resolve though as this thread was started by Sergio for feedback but consider my part of this resolved.

include/EventLoop.h
harryz2000: The timeout is 10 seconds. Isn't it?

bcostdolby: Yes sorry.

src/EventLoop.cpp
harryz2000: This is to allow the stop() function to be called within the thread's loop. As the thread would eventually exit after detach. I don't see in current code, the shutdown thread could be itself. We could use SetStopping() potentially. But I don't see big issue in this way either. If another thread is calling it, thread.join() would be called.

src/SystemPoll.cpp
harryz2000: sure

bcostdolby (approved):
lgtm

include/rtmp/rtmpconnection.h
```diff
@@ -37,15 +45,13 @@ class RTMPConnection :
 			RTMPNetConnection::Listener *listener,
 			std::function<void(bool)> accept
 		) = 0;
-		virtual void onDisconnect(const RTMPConnection::shared& con) = 0;
+		virtual void onDisconnect(RTMPConnection* con) = 0;
```
murillo128, line 48: why this change?

src/rtmp/rtmpserver.cpp
```diff
@@ -228,12 +228,23 @@ void RTMPServer::DeleteAllConnections()
...
 
 	//For all connections
-	for (auto &connection : connections)
-		//Stop it
-		connection->Stop();
-
-	//Clear all connections
-	connections.clear();
+	auto con = connections.begin();
+	while (con != connections.end())
```
murillo128, line 232: this is risky as we are not waiting for all the connections to end, so `onDisconnect()` may be called when the server object is deleted already

include/rtmp/rtmpconnection.h
harryz2000: The reason is if the RTMPConnection was freed and the thread is not stopped. The enable_shared_from_this() would be called for this callback while RTMPConnection itself was destructed. The the enable_shared_from_this() would cause bad weak ptr exception and segment fault. I believe this issue was also in old code. But it is rare that happens as we normally would stop the thread before free the RTMPConnection object.

src/rtmp/rtmpserver.cpp
harryz2000: Good point. Ideally we should refactor the RTMPServer class as well to use EvenLoop, which might help with this. For now, as we are using locks, we have to very careful to avoid dead lock. One solution is remove the listener in DeleteAllConnections to avoid callback as we would have already removed the connection.

harryz2000: I changed to code to clear listener while the RTMPServer is stopping. The DeleteAllConnections would now also wait until all RTMPConnections are stopped.

harryz2000:
I had to use a lock for accessing listener of RTMPConnection to avoid race condition. Fortunately the lock would be only called at connected or disconnected event so it wouldn't impact performance. I tried Sync(), but it is very risky. i.e. it would block if the event loop is not running.

I tested the changes. It works well.

include/rtmp/rtmpconnection.h
murillo128: >The reason is if the RTMPConnection was freed and the thread is not stopped.
How that could happen? You mean in the case we stop the thread on the RTMPConnection destructor

murillo128 (approved):
lgtm

include/rtmp/rtmpconnection.h
harryz2000: yes. When the stop() is called in destructor, in that case, enable_shared_from_this() cannot be called.

PR #383, opened 2024-11-07
feat: Import ArrayView from libwebrtc code

murillo128:


murillo128:
yes, not sure if it has more goodies

mildsunrise:
I don't see any extra goodies... if we could instead upgrade to c++20 that would be nice :point_right: :point_left: otherwise if it's too much work then we can merge this

murillo128:
fine to fastforward to c++20 if it is supported in all the pipelines

PR #384, opened 2024-11-07
refactor(h264): Avoid callbacks and return vector of nal units instead

murillo128:


mildsunrise (approved):
lgtm, i'm okay with the change

PR #386, opened 2024-11-08
rtp packetization refactor and unit test

yezhan10:


include/MediaFrameListenerBridge.h
```diff
@@ -85,9 +86,10 @@ class MediaFrameListenerBridge :
 	void Dispatch(std::chrono::milliseconds now);
 	void Update(std::chrono::milliseconds now);
 	void DispatchPackets(const std::vector<RTPPacket::shared>& packet);
-	void Packetize(const std::shared_ptr<MediaFrame>& frame, std::chrono::milliseconds scheduled);
-	void UpdateFrameInfoAndStats(const std::shared_ptr<MediaFrame>& frame, std::chrono::milliseconds now, std::chrono::milliseconds scheduled);
-	
+
+	void UpdateMediaStats(const std::shared_ptr<MediaFrame>& frame, uint32_t packetsCount, std::chrono::milliseconds now, std::chrono::milliseconds scheduled);
+	RTPPacketizer::FrameInfo RetrieveFrameInfo(const std::shared_ptr<MediaFrame>& frame);
```
murillo128, line 91: this info is duplicated from MediaFrame, why do we need it?

include/rtp/RTPPacketizer.h
```diff
@@ -3,10 +3,117 @@
...
+	struct FrameInfo
+	{
+		DWORD frameDuration;
+		DWORD startExtSeqNum;
+		DWORD clockRate;
+		DWORD extTimestamp;
+		std::optional<VideoFrameInfo> videoInfo;
+	};
+
+	static std::vector<RTPPacket::shared> GenerateRTPPackets(const std::shared_ptr<MediaFrame>& frame, BYTE codec)
```
murillo128, line 32: move implemention to .cpp

include/rtp/RTPPacketizer.h
```diff
@@ -25,6 +132,53 @@ class RTPPacketizer
 protected:
 	MediaFrame::Type	mediaType;
 	DWORD				codec;
+private:
+	static size_t CalculateAggregationEndIndex(const std::shared_ptr<MediaFrame>& frame, size_t startIdx, uint32_t maxSize)
```
murillo128, line 136: use `const MediaFrame::RtpPacketizationInfo& info` instead of `frame` as input

murillo128: if we are going to extract this function, let's move it to an h264xxx.h file instead

include/rtp/RTPPacketizer.h
```diff
@@ -25,6 +132,53 @@ class RTPPacketizer
...
+			//Check if we can aggregate in a single packet and it is not a fragmented packet
+			if (rtp.GetPrefixLen() || length > maxSize)
+					break;
+			//Include this packet in the stap-a 
+			last = i;
+		}
+		return last;
+	}
+
+	static void AggregatePayloads(const std::shared_ptr<MediaFrame>& frame, RTPPacket::shared& packet, size_t stapAStartIdx, size_t stapAEndIdx)
```
murillo128, line 158: same as above

murillo128 (changes requested):

include/MediaFrameListenerBridge.h
yezhan10: All frame-related info can be extracted from MediaFrame, however, I chose to store it since I wanted to avoid casting mediaFrame to audioFrame or videoFrame multiple times.

include/rtp/RTPPacketizer.h
yezhan10: Not sure if I prefer to move it to h264Packetizer.h since `CalculateAggregationEndIndex()` and `AggregatePayloads()` are not h264 particular. The function would be the same regardless of the codec type, so it makes less sense to me to move them to a file that pertaining to a specific codec.

include/MediaFrameListenerBridge.h
yezhan10: I can delete this function, and move frame info extraction logic inside `SetPacketProps()` if you prefer this way. I don't have strong preference.

src/rtp/RTPPacketizer.cpp
```diff
@@ -0,0 +1,159 @@
...
+			break;
+		}
+		default:
+			break;
+		
+	}
+
+	for (size_t i=0; i<rtpPackets.size(); i++)
+	{
+		auto packet = rtpPackets[i];
```
murillo128, line 135: Do not copy, it will cause a new shared pointer to be created, degrading performance
```suggestion
		auto& packet = rtpPackets[i];
```

src/rtp/RTPPacketizer.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+		rtpPackets.push_back(std::move(packet));
+	}
+	return rtpPackets;
+}
+
+void RTPPacketizer::SetPacketsProp(const std::shared_ptr<MediaFrame>& frame, const std::vector<RTPPacket::shared>& rtpPackets, uint64_t lastTimestamp, uint64_t extSeqNum, uint32_t targetBitrateHint)
+{
+	uint32_t rate = 0;
+	RTPPacketizer::VideoFrameInfo videoInfo {0};
```
murillo128, line 100: I don't see the value of having a struct at packetizer level when we are using only internally in this function

murillo128 (changes requested):

include/rtp/RTPPacketizer.h
```diff
@@ -25,6 +40,9 @@ class RTPPacketizer
 protected:
 	MediaFrame::Type	mediaType;
 	DWORD				codec;
+private:
+	static size_t CalculateAggregationEndIndex(const MediaFrame::RtpPacketizationInfo& info, size_t startIdx, uint32_t maxSize);
```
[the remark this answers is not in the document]
yezhan10: I replied to your previous comment about moving the functions to h264. I think that these two functions are not codec specific operations, so it makes more sense to not move them to a codec specific file?

include/rtp/RTPPacketizer.h
murillo128: they make no sense for codecs other than h264

include/rtp/RTPPacketizer.h
```diff
@@ -3,10 +3,25 @@
...
+	struct VideoFrameInfo
+	{
+		bool isIntra;
+		uint16_t width;
+		uint16_t height;
+		DWORD targetBitrate;
+		DWORD targetFps;
+	};
+
+	static std::vector<RTPPacket::shared> GenerateRTPPackets(const std::shared_ptr<MediaFrame>& frame, BYTE codec);
```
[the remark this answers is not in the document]
yezhan10: yeah i chose to pass it as param since codec is not a member of MediaFrame, so I have to cast it again inside this function if I choose not to pass the codec. And codec is already member variable of MediaFrameListenerBridge.

include/rtp/RTPPacketizer.h
```diff
@@ -3,10 +3,25 @@
 #include "config.h"
 #include "media.h"
 #include "rtp/RTPPacket.h"
+#include "VideoLayerSelector.h"
 
 class RTPPacketizer
 {
 public:
+	static constexpr uint32_t StapANaluSize = 2;
+	static constexpr uint8_t StapAHeader = 0b0'11'11000; // STAP-A header nalu type = 24
```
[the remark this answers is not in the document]
yezhan10: Not sure I am following, does not belong to this class? What's the intended usage of this class? I thought that this class is responsible for packetize stapA regardless the codec type? You want it to be moved to h26xPacketizer?

include/rtp/RTPPacketizer.h
yezhan10: ok, i thought that stap-a can be applied to other codec as well. We just support stapA for h264 now. If this is the case, I will move it to h264packetizer instead

src/MediaFrameListenerBridge.cpp
```diff
@@ -297,190 +296,9 @@ void  MediaFrameListenerBridge::onMediaFrameAsync(std::chrono::milliseconds now,
...
-}
-
-void MediaFrameListenerBridge::UpdateFrameInfoAndStats(const std::shared_ptr<MediaFrame>& frame, std::chrono::milliseconds now, std::chrono::milliseconds scheduled)
-{
 	type = frame->GetType();
-	uint32_t frameSize = 0;
-	uint64_t time = frame->GetTime();
-	uint64_t ms = scheduled.count();
+	lastTimestamp = baseTimestamp + (frame->GetTimeStamp()-firstTimestamp);
 	switch(type)
```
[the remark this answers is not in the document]
yezhan10: i think so, i got impression that the codec is a member variable and it's used somewhere else in MediaFrameListenerBrige, but seems like this is not the case. I will remove it

include/rtp/RTPPacketizer.h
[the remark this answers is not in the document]
murillo128: rtpacketizer is codec agnostic, so it doesn't make sense to have any h264 codec specific info here

src/MediaFrameListenerBridge.cpp
```diff
@@ -297,244 +296,73 @@ void  MediaFrameListenerBridge::onMediaFrameAsync(std::chrono::milliseconds now,
...
-			bool isPFrame = !(videoFrame->IsIntra() || videoFrame->IsBFrame());
-			accumulatorPFrames.Update(now.count(), isPFrame ? 1 : 0);
 			break;
 		}
 		default:
-			return;
+			break;
+	}
+	// Create RTPPackets for this frame and set properties of all generated packets
+	auto rtpPackets = RTPPacketizer::GenerateRTPPackets(frame);
```
murillo128, line 336: I think that both methods should be merged together

src/h264/H264Packetizer.h
```diff
@@ -13,7 +13,14 @@
 class H264Packetizer : public H26xPacketizer
 {
 public:
+	static constexpr uint32_t StapANaluSize = 2;
```
murillo128, line 16: move to the .cpp

murillo128 (changes requested):
just minor changes, almost lgtm

src/MediaFrameListenerBridge.cpp
yezhan10: which two methods? GenerateRTPPackets and SetPacketsProp?

src/h264/H264Packetizer.h
murillo128: yes

PR #388, opened 2024-11-14
fix: [DIOS-7648] Handle EINTR by retrying syscall through main loop

bcostdolby:


src/SystemPoll.cpp
```diff
@@ -185,8 +185,16 @@ Poll::WaitResult SystemPoll::Wait(uint32_t timeOutMs)
...
 	if (ret < 0)
 	{
-		Error("-SystemPoll::Wait() | poll() error. error: %s\n", strerror(errno));
-		return Poll::WaitResult::Error;
+		// If this thread happened to wakeup because it handled a signal that was delivered
+		// here then we want to just re-enter the main loop
+		// and try again (after potentially checking exit status etc)
+		//
+		// Instead of early return we can safely check the events below
+		if (errno != EINTR)
```
eck1001, line 193: nitpick: personally, I like flattening if statements when we can - what do you think?

`if (ret < 0 && errno != EINTR)`

eck1001: my rationale is that this helps reduce the upstream context that is required + less blocks of code.

src/rtmp/rtmpserver.cpp
```diff
@@ -159,6 +163,11 @@ int RTMPServer::Run()
 
 		//Accpept incoming connections
 		int fd = accept(server,NULL,0);
+		while (fd < 0 && errno == EINTR)
+		{
+			UltraDebug("EINTR during accept trying again\n");
+			fd = accept(server,NULL,0);
```
harryz2000, line 169: Is that possible it keeps returns EINTR and we end up of infinite loop? If that case, it might block the server.

mildsunrise: no, eintr should be sporadic afaik

mildsunrise (approved):
lgtm, but maybe we could make the threads not receive the signals to begin with?

src/rtmp/rtmpserver.cpp
harryz2000: OK. If the error is temporary. All good.

bcostdolby:
> lgtm, but maybe we could make the threads not receive the signals to begin with?

I considered this as well, but in this case I believe some of the signals are being delivered from outside the environment (emulated docker container as on a m3 mac) and I wasn't 100% certain it would resolve the issue. Also it is very common to handle EINTR in this way manually. 

If we think adding sigmasks to the various threads is helpful we can do that as well but would target another ticket.

harryz2000 (approved):
lgtm

PR #389, opened 2024-11-14
fix: [DIOS-7649] Add SOREUSEPORT option to bound TCP sockets to workon BSD systems

bcostdolby:
…on BSD


src/ws/websocketserver.cpp
```diff
@@ -92,7 +92,16 @@ int WebSocketServer::Run()
 
 	//Set SO_REUSEADDR on a socket to true (1):
 	int optval = 1;
-	setsockopt(server, SOL_SOCKET, SO_REUSEADDR, &optval, sizeof(optval));
+	int result = setsockopt(server, SOL_SOCKET, SO_REUSEADDR, &optval, sizeof(optval));
```
eck1001, line 95: change-request: Lines 94-97 were these suppose to be deleted? They look like duplicates.

eck1001 (changes requested):

src/rtmp/rtmpserver.cpp
```diff
@@ -76,7 +76,20 @@ int RTMPServer::BindServer()
...
 // Ignore coverity error: "this->server" is passed to a parameter that cannot be negative.
 // coverity[negative_returns]
-	(void)setsockopt(server, SOL_SOCKET, SO_REUSEADDR, &optval, sizeof(optval));
+	int result = setsockopt(server, SOL_SOCKET, SO_REUSEADDR, &optval, sizeof(optval));
+	if (result < 0)
+	{
+		Warning("-RTMPServer::BindServer() Failed to set SO_REUSEADDR on socket. reason: %s\n", strerror(errno));
+	}
+
+	// See https://stackoverflow.com/questions/14388706/how-do-so-reuseaddr-and-so-reuseport-differ
```
eck1001, line 85: 🎉: love the context notes -- thanks - BSD sockets... what a world

harryz2000:
lgtm

harryz2000 (approved):

mildsunrise:
wait, do we want to make medooze work in systems other than Linux? we abandoned that path quite some time ago, and I'm not sure it's a good idea to retake it...

harryz2000:
I think this is for Linux over Docker. @bcostdolby can confirm.

bcostdolby:
> wait, do we want to make medooze work in systems other than Linux? we abandoned that path quite some time ago, and I'm not sure it's a good idea to retake it...

I am just trying to make it work in docker on the m3. It is using rosetta to emulate so is not a arm arch running but intel. It works fine but there are some minor behaviour differences and IMO using SO_REUSEPORT is correct anyway.

mildsunrise:
I am still a bit confused, doesn't docker desktop start a Linux VM and run docker deamon there? why would this change be needed for such a setup, if it's running inside a Linux VM? :o

eck1001:
Docker linux containers still share the host kernel and SO_REUSEPORT is a feature implemented at the kernel level

